PR #5, opened 2017-04-04
get all patients from backend and show patient count

beaugogh:


src/app/modules/data-selection/components/patient-selection/patient-selection.component.ts
```diff
@@ -1,25 +1,33 @@
...
 })
 export class PatientSelectionComponent implements OnInit {
+  patientCount: number;
 
-  constructor(private resourceService: ResourceService) { }
+  constructor(private resourceService: ResourceService, private workflowService: WorkflowService) {
+  }
 
   ngOnInit() {
+    this.patientCount = 0;
```
kretep: Couldn't this be done in the constructor?

src/app/modules/data-selection/components/patient-selection/patient-selection.component.ts
```diff
@@ -1,25 +1,33 @@
...
     this.resourceService.getPatients().subscribe(
-      patients => {
-        console.log('patients: ', patients);
+      patientsObj => {
+        console.log('patients: ', patientsObj);
+        currentWorkflow.setPatients(patientsObj['patients']);
+        this.patientCount = currentWorkflow.getPatients().length;
       },
       err => {
         console.log(err);
```
OlafMeuwese: Maybe console.warn or console.error is preferred when logging an error instead of console.log

beaugogh: yeah, we are still in an experimental stage, due to the incremental nature of the tickets. Eventually all the console.xx would be removed

PR #8, opened 2017-04-05
let workflow store accordion states and refactor service injection in…

beaugogh:
… app comp


src/app/modules/shared/services/workflow.service.ts
```diff
@@ -3,11 +3,13 @@ import {Workflow} from "../models/workflow";
 
 @Injectable()
 export class WorkflowService {
+  public what: string[];
```
kretep: what? I don't find this name very descriptive

PR #12, opened 2017-04-11
add hierachical constraint

beaugogh:


beaugogh:
Added the hierarchy structure: each specified constraint model (such as study-constraint, concept-constraint) extends from a common constraint model, each specified constraint visual component also extends from a common constraint component. and each constraint component contains exactly one constraint model. The constraint model stores the query-related information, the component is only responsible for rendering this information.

src/app/modules/data-selection/constraint-components/constraint/constraint.component.ts
```diff
@@ -0,0 +1,22 @@
...
+export class ConstraintComponent implements OnInit {
+  @Input() constraint: Constraint;
+
+  constructor() {
+  }
+
+  ngOnInit() {
+  }
+
+  removeSelfComponent() {
```
kretep: What will this do? Maybe add docs?

src/app/modules/shared/models/constraints/combination-constraint.ts
```diff
@@ -0,0 +1,33 @@
+import {Constraint} from './constraint';
+
+export class CombinationConstraint implements Constraint {
+
+  private children: Constraint[];
+  public NOT: boolean;
+  public AND: boolean;
+  public OR: boolean;
```
kretep: As these attributes are not constants, I'd rather not use all-caps for the name. Also, aren't AND and OR mutually exclusive? And as they are booleans, maybe isNot and isAnd would be better names, and the OR would be deduced from isAnd.

src/app/modules/data-selection/constraint-components/constraint/constraint.component.html
```diff
@@ -0,0 +1,8 @@
+<div>
+  <div *ngIf="constraint.getConstraintType() == 'combination-constraint'">
```
kretep: Couldn't this be done with instanceof? Also, it may be nicer if we could instantiate the constraint components on the fly, instead of hard-coding them in a switch-type check.

PR #37, opened 2017-05-03
	finalize categorical concept constraint

beaugogh:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "glowing-bear",
   "version": "0.0.0",
-  "license": "MIT",
+  "license": "GPLv3",
```
kretep: I think the proper identifier is GPL-3.0, but we are using GPL-3.0+, right?

kretep: See https://spdx.org/licenses/

PR #38, opened 2017-05-04
Dev

beaugogh:


beaugogh:
the env.json still needs to be updated, now it is 'env':'dev' whereas we need 'prod' for the master branch. not sure why env.json still got picked up by git, since it should be ignored

PR #40, opened 2017-05-09
split inclusion and exclusion, remove not/and/or

beaugogh:


src/app/modules/data-selection/constraint-components/combination-constraint/combination-constraint.component.html
```diff
@@ -1,27 +1,13 @@
...
-    <label class="form-check-label">
-      <input [(ngModel)]="isNot" class="form-check-input" type="checkbox">
-      NOT
-    </label>
+  <!-- Child Constraints -->
+  <constraint *ngFor="let child of children"
+              [constraint]="child"
+              (constraintRemoved)="onConstraintRemoved(child)"></constraint>
 
-    <div class="btn-group">
```
kretep: I think the idea was to replace the buttons with a different way to control and/or, but it seems this just completely removes them.

beaugogh: see the latest commit

PR #42, opened 2017-05-10
Add date constraint to concept

kretep:


src/app/modules/shared/models/constraints/concept-constraint.ts
```diff
@@ -1,12 +1,17 @@
...
 
 export class ConceptConstraint implements Constraint {
 
   private _type: string;
   private _concept:Concept;
   private _values: Value[];
+  applyDateConstraint: boolean = false;
+  dateOperator: DateOperatorState = DateOperatorState.BETWEEN;
+  date1: Date = new Date();
+  date2: Date = new Date();
```
beaugogh: you mean startDate and endDate?

src/app/modules/shared/models/constraints/concept-constraint.ts
```diff
@@ -63,4 +72,43 @@ export class ConceptConstraint implements Constraint {
     }
     return 'Concept';
   }
+
+  buildDateConstraint():Object {
```
beaugogh: this type of logic should be put into constraint-service, no?

src/app/modules/shared/models/constraints/concept-constraint.ts
kretep: Not necessarily, as they can be either the limits of a range or a reference date, depending on the operator. I also wanted to avoid confusion with the observation's start and end date.

src/app/modules/shared/models/constraints/concept-constraint.ts
kretep: Yes, I'll move it there

PR #49, opened 2017-05-15
tree on the side

beaugogh:


beaugogh:
@kretep do not merge this yet, since it is not needed in the demo today

kretep:
@beaugogh can't you just omit the env.json from all the commits?

PR #54, opened 2017-05-17
Show spinner on loading and animation on completion

kretep:


src/app/modules/shared/services/constraint.service.ts
```diff
@@ -25,14 +31,20 @@ export class ConstraintService {
   }
 
   updatePatients() {
+    this.loadingStateInclusion = "loading";
```
beaugogh: create a enum for loading state?

kretep: The problem is that the string values are also used for the animation state. But I now use a string literal type (LoadingState), to prevent it from having an invalid value.

PR #81, opened 2017-05-30
multiple auto-complete style fix

beaugogh:
see ticket https://jira.thehyve.nl/browse/GBD-92


src/app/modules/data-selection/constraint-components/constraint/constraint.component.ts
```diff
@@ -76,8 +76,15 @@ export class ConstraintComponent implements OnInit {
       else if (this.constraint.getClassName() === droppedConstraint.getClassName()) {
         if (this.constraint instanceof StudyConstraint) {
           let study = (<StudyConstraint>droppedConstraint).studies[0];
-          (<StudyConstraint>this.constraint).studies.push(study);
-          this.constraintService.update();
+          let studies = (<StudyConstraint>this.constraint).studies;
+          let studyIds = [];
+          for(let _study of studies) {
```
kretep: Maybe something like this:

```
let studies = (<StudyConstraint>this.constraint).studies;
studies.filter(item => item.studyId == study.studyId);
if (studyIds.length == 0) { ...
```

Or at least put a comment at this part so we know what the code is for.

PR #194, opened 2017-10-10
Gradle publish

gijskant:
- Add package/publish script (gradle)
- Fix failing tests on Travis (due to jasime version).


beaugogh:
strange now it passes while yesterday it failed

PR #251, opened 2017-12-19
Limit backend calls

gijskant:
- Move formatting logic to separate helper class.
- Remove duplicate loading state enum type.
- Refactor query counts functionality:
  - mark state as dirty after modifying constraints or variable selection;
  - display waiting indicators when retrieving counts;
  - split long functions;
- Fix bug with 'select all' (adds to instead of overrides current selection)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,7 @@ buildscript {
 }
 
 ext {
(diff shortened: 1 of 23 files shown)


beaugogh:
added the corresponding functionalities in this PR in another branch, while keeping the UI inlined with existing requirements

PR #270, opened 2018-02-08
Create patient set before fetching tree counts for patient selection.

gijskant:
Save the subject selection of step 1 automatically as subject set in the backend. That subject set is used as the subject constraint for counts in step 2 and for export.

### DIFF
--- a/src/app/services/query.service.ts
+++ b/src/app/services/query.service.ts
@@ -444,30 +444,6 @@ export class QueryService {
      */
   }
 
-  // private updatePatientSet(timeStamp) {
-  //   const selectionConstraint = this.constraintService.generateSelectionConstraint();
-  //   this.resourceService.createPatientSet(name, selectionConstraint)
-  //     .subscribe(
-  //       patientSetObj => {
-  //         this.patientSet_1 = new PatientSetConstraint();
-  //         this.patientSet_1.setSize = patientSetObj['setSize'];
-  //         this.patientSet_1.id = patientSetObj['id'];
-  //         this.patientSet_1.status = patientSetObj['status'];
-  //         /*
-  //         * update concept and study counts in the first step
-  //         */
-  //         this.resourceService.getCountsPerStudyAndConcept(this.patientSet_1)
-  //           .subscribe(
-  //             (countObj) => {
-  //               this.updateConceptsAndStudies(timeStamp);
-  //             },
-  //             err => this.handle_error(err)
-  //           );
-  //       },
-  //       err => this.handle_error(err)
-  //     );
-  // }
-
   /**
    * ------------------------------------------------- END: step 1 ---------------------------------------------------
    */
--- a/src/app/services/query.service.ts
+++ b/src/app/services/query.service.ts
@@ -7,6 +7,9 @@ import {ConstraintService} from './constraint.service';
 import {Step} from '../models/step';
 import {PatientSetConstraint} from '../models/constraints/patient-set-constraint';
 import {FormatHelper} from '../utilities/FormatHelper';
+import {PatientSetResponse} from '../models/patient-set-response';
+import {Constraint} from '../models/constraints/constraint';
+import {AppConfig} from '../config/app.config';
 
 type LoadingState = 'loading' | 'complete';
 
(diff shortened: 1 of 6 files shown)


src/app/config/config.dev.json
```diff
@@ -1,5 +1,7 @@
 {
   "api-url": "http://localhost:8080",
   "api-version": "v2",
-  "app-url": "http://localhost:4200"
+  "app-url": "http://localhost:4200",
+  "tree-node-counts-update": true,
```
beaugogh: flags such as tree-node-counts-update have been moved to query.service, as we discussed earlier

src/app/config/config.dev.json
```diff
@@ -1,5 +1,7 @@
...
+  "autosave-subject-sets": false
```
beaugogh: I would also put this new flag to the same place as tree-node-counts-update, i.e. in query.service constructor

src/app/config/config.prod.json
```diff
@@ -1,5 +1,7 @@
 {
   "api-url": "https://transmart-test.thehyve.net",
   "api-version": "v2",
-  "app-url": "https://glowingbear2-head.thehyve.net"
+  "app-url": "https://glowingbear2-head.thehyve.net",
+  "tree-node-counts-update": true,
+  "autosave-subject-sets": false
```
beaugogh: see my comments on config.dev.json

src/app/config/config.dev.json
gijskant: Yes, but then you cannot set them from outside, e.g, using Puppet. Default values can be in query.service, but you should be able to read the values from the json if present.

gijskant: The way I implemented it now allows you to omit these fields in the json, so puppet only needs to set them when they need to be overridden.

beaugogh: If they can be omitted, then they should be omitted here, right?

beaugogh: I mean we can still try to read the flags in query.service, since puppet might override config.prod.json some day, but since this is glowingbear code, we keep the config jsons unchanged

gijskant: I thought they could serve as examples, with the default values filled in. But I could also document the possible values in the readme instead.

src/app/services/query.service.ts
```diff
@@ -134,15 +139,22 @@ export class QueryService {
...
+  constructor(private appConfig: AppConfig,
+              private resourceService: ResourceService,
               private treeNodeService: TreeNodeService,
               private constraintService: ConstraintService) {
     this.instantCountsUpdate_1 = false;
     this.instantCountsUpdate_2 = false;
-    this.treeNodeCountsUpdate = true;
+    this.treeNodeCountsUpdate = appConfig.getConfig('tree-node-counts-update', true);
     this.countsRelay = false;
+    this.autosaveSubjectSets = appConfig.getConfig('autosave-subject-sets', false);
```
beaugogh: just a remark about the way that glowing bear is handling these parameters/flags now, there is this inconsistency: some flags are directly set in query.service, such as instantCountsUpdate_1, some flags are set via external config file such as treeNodeCountsUpdate, which may need some improvement. 

ps: omg, almost forgot that I am on holiday! :)

src/app/services/query.service.ts
```diff
@@ -180,7 +192,23 @@ export class QueryService {
...
 
-  /**
+  private mergeInclusionAndExclusionCounts(initialUpdate?: boolean) {
+    if (this.autosaveSubjectSets) {
+      // Not computing the counts based on inclusion and exclusion counts,
+      // but using the subject set and counts per study concept response instead.
+      return;
+    }
+    this.subjectCount_1 = this.inclusionSubjectCount - this.exclusionSubjectCount;
+    this.observationCount_1 = this.inclusionObservationCount - this.exclusionObservationCount;
```
ewelinagr: so if inclusionObservationCount < exclusionObservationCount, we will have observationCount < 0? Or it is not possible?

gijskant: It is not possible by construction: the exclusion constraint is of the form (inclusionConstraint AND NOT (...)), so will always be maximum the number of elements of the inclusion result.

gijskant:
Replaced by a PR to master: #272.

PR #271, opened 2018-02-09
Clear default prod config; change default environments.

gijskant:
- Change the default config environment to `dev` for development. `ng serve` will use `localhost:8080` as back end and `localhost:4200` for the front end by default.
- Change the default config environment to 'prod' for production/deployment use.
- Change the default `api-url` and `app-url` to an invalid url (`change me`) to force overriding these values at deployment time.
- Clean up a `switch` statement in `AppConfig`.


mickmis:
It works, but is there a reason why both app/config/env*.json and environments/environment*.json are used? Looks like they are doing the same thing twice.

mickmis:
Gijs clarified this: the environments/environment*.ts files get minified when built in production mode, while the app/config/env*.json can still be modified.

PR #274, opened 2018-02-14
Add buttons to cancel and archive export jobs

ewelinagr:
NTRDEV-294, NTRDEV-314

### DIFF
--- a/src/app/models/export-job.ts
+++ b/src/app/models/export-job.ts
@@ -5,6 +5,7 @@ export class ExportJob {
   private _jobStatusTime: string;
   private _userId: string;
   private _viewerURL: string;
+  private _isInDisabledState: boolean;
 
   get id(): string {
     return this._id;
@@ -52,5 +53,13 @@ export class ExportJob {
 
   set viewerURL(value: string) {
     this._viewerURL = value;
+  }
+
+  get isInDisabledState(): boolean {
+    return this._isInDisabledState;
+  }
+
+  set isInDisabledState(value: boolean) {
+    this._isInDisabledState = value;
   }
 }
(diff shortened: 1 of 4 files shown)


src/app/models/export-job.ts
```diff
@@ -53,4 +54,12 @@ export class ExportJob {
   set viewerURL(value: string) {
     this._viewerURL = value;
   }
+
+  get isInDisabledState(): boolean {
+    return this._isInDisabledState;
```
gijskant: Why not simply `_disabled` with getter `isDisabled` and setter `setDisabled`?

src/app/modules/gb-data-selection-module/accordion-components/gb-export/gb-export.component.html
```diff
@@ -74,20 +74,35 @@ <h5>Previous Exports</h5>
...
                 *ngIf="job.jobStatus === 'Completed'"
                 class="btn btn-outline-primary btn-sm"
+                [disabled] = "job.isInDisabledState"
                 (click)="downloadExportJob(job)">
           <span class="fa fa-download"></span>
           Download
         </button>
-
+        <button type="button"
+                *ngIf="['Completed','Error','Cancelled'].indexOf(job.jobStatus) < 0;else archive"
```
gijskant: You can use `includes` instead of `indexOf`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

src/app/modules/gb-data-selection-module/accordion-components/gb-export/gb-export.component.ts
```diff
@@ -136,6 +138,28 @@ export class GbExportComponent implements OnInit {
...
+  }
+
+  archiveExportJob(job) {
+    job.isInDisabledState = true;
+    this.resourceService.archiveExportJob(job.id)
+      .subscribe(
+        response => {
+          this.updateExportJobs();
+        },
+        err => console.error(err)
```
gijskant: I see that poor error handling is currently standard in Glowing Bear. We should fix that at some point.

PR #277, opened 2018-02-14
Add a checkbox to choose for export of the datums

ewelinagr:
NTRDEV-310


ewelinagr:
Back-end part is implemented [here](https://github.com/thehyve/transmart-core/pull/165)

PR #280, opened 2018-02-21
Fix subject id import for dos/windows line breaks.

gijskant:
Fix issue with import of subject ids that contains Windows line breaks (`\r\n`).
Also skip empty lines.

Fixes [NTRDEV-320](https://jira.thehyve.nl/secure/RapidBoard.jspa?rapidView=110&view=detail&selectedIssue=NTRDEV-320).

### DIFF
--- a/src/app/modules/gb-side-panel-module/accordion-components/gb-queries/gb-queries.component.ts
+++ b/src/app/modules/gb-side-panel-module/accordion-components/gb-queries/gb-queries.component.ts
@@ -37,7 +37,7 @@ export class GbQueriesComponent implements OnInit {
 
   processSubjectIdsUpload(fileContents: string, fileName: string): Query {
     // we assume the text contains a list of subject Ids
-    let subjectIds: string[] = fileContents.split(/(\r\n)+/)
+    let subjectIds: string[] = fileContents.split(/(\r?\n)+/)
       .map(id => id.trim())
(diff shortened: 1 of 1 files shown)


src/app/modules/gb-side-panel-module/accordion-components/gb-queries/gb-queries.component.ts
```diff
@@ -35,12 +35,27 @@ export class GbQueriesComponent implements OnInit {
     uploadElm.click();
   }
 
+  processSubjectIdsUpload(fileContents: string, fileName: string): Query {
+    // we assume the text contains a list of subject Ids
+    let subjectIds: string[] = fileContents.split(/(\r\n)+/)
```
ewelinagr: maybe better to use `split(/\r?\n/)`? It seems like `split(/(\r\n)+/)` is not interpreting correctly new lines in Unix systems - I tested something like ` "  hello  \n\nfoo \n bar\nworld  \n"`, got `Array [ "hello  \n\nfoo \n bar\nworld" ]`

gijskant: I'll fix it

gijskant: And now conflicting changes....

PR #292, opened 2018-03-06
Change query import, add criteria imports in step 1 and 2.

ewelinagr:
NTRDEV-324

### DIFF
--- a/src/app/modules/gb-data-selection-module/accordion-components/gb-projection/gb-projection.component.html
+++ b/src/app/modules/gb-data-selection-module/accordion-components/gb-projection/gb-projection.component.html
@@ -1,4 +1,12 @@
 <div>
+  <div class="float-right" style="margin-right: 2em">
+    <input id="step2CriteriaFileUpload"
+           type="file"
+           style="display: none"/>
+    <button type="button" (click)="importCriteria()"
+            class="btn btn-primary btn-sm gb-data-selection-control-btn float-right">Import Criteria
+    </button>
+  </div>
   <div class="container form-inline">
     <button type="button" (click)="expandAll(true)"
             class="btn btn-primary btn-sm gb-data-selection-control-btn">Expand All</button>
@@ -11,8 +19,7 @@
     <span>&nbsp;</span>
     <button type="button" (click)="checkAll(false)"
             class="btn btn-primary btn-sm gb-data-selection-control-btn">Check None</button>
-  </div>
-
+</div>
   <p-tree [value]="projectionTreeData"
           selectionMode="checkbox"
           (onNodeSelect)="updateCounts()"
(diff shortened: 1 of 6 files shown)


src/app/modules/gb-data-selection-module/accordion-components/gb-projection/gb-projection.component.html
```diff
@@ -1,4 +1,12 @@
 <div>
+  <div class="float-right" style="margin-right: 2em">
```
beaugogh: align the import criteria button with the other buttons ?

src/app/services/query.service.ts
```diff
@@ -597,6 +597,162 @@ export class QueryService {
       );
   }
 
+  /**
+   * Import query from a file
+   * @param event
+   */
+  public importQuery(event) {
```
beaugogh: I think the query import, step1 import and step2 import are different, i.e. the formats of the queries imported in these three places are different. I would put the import file logic into the respective component controllers, as they are DOM element interaction, and keep the query.service more about count update and manipulation of parsed queries.

src/app/modules/gb-data-selection-module/accordion-components/gb-selection/gb-selection.component.ts
```diff
@@ -69,6 +72,62 @@ export class GbSelectionComponent implements OnInit {
...
+          'type': 'patient_set',
+          'subjectIds': data.split('\n')
+        },
+        'observationsQuery': {}
+      };
+      return query;
+    } else if (file.type === 'application/json') {
+      let _json = JSON.parse(data);
+      // If the json is of standard format
+      if (_json['patientsQuery'] || _json['observationsQuery']) {
```
beaugogh: we probably need to limit ourselves about which json format to accept in importing criteria into step 1: query import is capable of importing both patient and observation criteria, step 1 is only capable of importing patient query, and step 2 is only capably of importing observation query.

beaugogh: e.g. for step 1 import, we only accept this json format: 
`"patientsQuery": {
    "type": "true"
  }`

beaugogh: for step 2 import, we only parse the content in this json format:
`"observationsQuery": {
    "data": []
  }`

or 
`"names": []`

or 
`"paths": []`

ewelinagr: but I guess "observationsQuery" can still be a part of a file imported in step1, just it will be skipped. 
For example, if someone exported a query and want to import it to step 1, it would be possible without editing. The only part that would be used from the export file during the import would be "patientsQuery" part. 
What do you think, @beaugogh ?

beaugogh: yes, exactly!

PR #293, opened 2018-03-06
not updating inclusion counts when subject set autosaved

beaugogh:


src/app/services/query.service.ts
```diff
@@ -229,27 +228,31 @@ export class QueryService {
...
-            this.inclusionSubjectCount = countResponse['patientCount'];
-            this.inclusionObservationCount = countResponse['observationCount'];
-            this.loadingStateInclusion = 'complete';
-            if (this.loadingStateTotal !== 'complete' && this.loadingStateExclusion === 'complete') {
-              this.mergeInclusionAndExclusionCounts(initialUpdate);
-              // relay the current counts to the next step: subjects and observations
-              this.relayCounts_1_2();
+    // No need to compute the inclusion count if subject sets are auto saved,
+    // the saved subject sets will be used in updateConceptsAndStudies to calculate the inclusion counts
+    if (!this.autosaveSubjectSets) {
```
ewelinagr: Shouldn't the same check be added to updateExclusionCounts function (before this.resourceService.getCounts(exclusionConstraint))?

beaugogh: the exclusion count is always called, if there is no autosaved subject set, we call inclusion count, then compute the final counts based on inclusion and exclusion counts. If there is autosaved subject set, we use that to get the final counts first, then we calculate the inclusion counts.

ewelinagr: ah ok, so the exclusion count is always called, but mergeInclusionAndExclusionCounts  after the exclusion count
is done only, when there is no autosaved subject set?
I'm just wondering about the condition from line 198, if the loadingStateTotal will be always set to 'complete';

beaugogh: line 198? which one?

ewelinagr: if (!this.autosaveSubjectSets) inside queryService.mergeInclusionAndExclusionCounts

beaugogh: if we reach there, then it's completed, since both subject count and observation count have been updated

beaugogh: hmm, maybe the wording for loadingStateTotal is confusing, it is from legacy, and it actually means only for the first step. I will change that

beaugogh: @ewelinagr updated the loadingStateTotal flag to loadingStateTotal_1 !

PR #299, opened 2018-03-07
NTR Workflow 

beaugogh:


src/app/services/query.service.ts
```diff
@@ -128,7 +128,7 @@ export class QueryService {
   private _treeNodeCountsUpdate: boolean;
   private _exportFormats: object[] = [];
   private _isLoadingExportFormats = false;
-  private _exportDataView = 'default';
+  private _exportDataView = 'surveyTable'; // NTR specific
```
beaugogh: do we keep the NTR specific export data view format, or use 'default'?

beaugogh: Maybe keep it as 'default', since down below the _exportDataView value gets updated via config service

PR #303, opened 2018-03-14
add data table as 3rd step

beaugogh:


src/app/modules/gb-data-selection-module/gb-data-selection.component.html
```diff
@@ -65,18 +65,27 @@
...
+    <md2-accordion-tab>
+      <md2-accordion-header>
+        <div class="gb-data-selection-accordion-header">
+          <div class="gb-data-selection-accordion-sub-header-1">
+            Step 3: View and arrange the selected data
+          </div>
+          <div class="form-inline gb-data-selection-accordion-sub-header-2">
+            <div class="gb-data-selection-accordion-sub-header-3">
+              <span class="gb-data-selection-emphasis-text">{{subjectCount_2}} subjects</span>,
+              <span class="gb-data-selection-emphasis-text">{{observationCount_2}} observations</span>
```
ewelinagr: why these counts here?

beaugogh: when the user create or filter the table, there is a further selection of data, which results new counts

ewelinagr: ah ok, I thought we will have separate counts per each step so a count only includes update from the current step.

PR #304, opened 2018-03-18
 add dimension director to step 3

beaugogh:


src/app/models/table-models/dimension.ts
```diff
@@ -0,0 +1,25 @@
+export class Dimension {
```
ewelinagr: I thought of a bit different model for dimension - to include index, label, maybe elements as well (to match Jan's response format a bit). I'm not sure what does selected mean?

src/app/models/table-models/dimension-zone-enum.ts
```diff
@@ -0,0 +1,5 @@
+export enum DimensionZoneEnum {
+  Rows = 'Rows',
+  Columns = 'Columns',
+  Candidates = 'Candidates'
+}
```
ewelinagr: What exactly is included in Candidates? Does it mean all dimensions possible and then some of them are in rows, some columns and some nowhere or these are just the ones that are not assigned to rows and columns?

src/app/models/table-models/dimension.ts
beaugogh: yes, you can add extra fields to Dimension of course

beaugogh: 'selected' is a flag to indicate if the dimension is selected or not

src/app/models/table-models/dimension-zone-enum.ts
beaugogh: the candidates ware all the dimensions available from the previous step, the user will assign candidates to rows and columns to compose the table. And indeed, the user can pick a subset of candidates to assign to rows and columns

PR #306, opened 2018-03-19
Add saving and restoring of a data table state + refactoring of a resource service

ewelinagr:
TMT-83, TMT-84

### DIFF
--- a/src/app/models/query-models/query.ts
+++ b/src/app/models/query-models/query.ts
@@ -1,6 +1,6 @@
 import {QuerySubscriptionFrequency} from './query-subscription-frequency';
 import {QueryDiffRecord} from './query-diff-record';
-import {DataTableState} from "../table-models/data-table-state";
+import {TransmartTableState} from "../transmart-resource-models/transmart-table-state";
 
 export class Query {
 
@@ -26,8 +26,8 @@ export class Query {
   private _observationsQuery: {data: string[]};
   // The visual indicator flags the visibility of the query
   private _visible: boolean;
-  // Additional information about the query, e.g. a data table state
-  private _dataTableState: DataTableState;
+  // The information about the saved data table state
+  private _dataTableState: TransmartTableState;
 
   /*
    * Subscription feature
@@ -159,11 +159,11 @@ export class Query {
     this._visible = value;
   }
 
-  get dataTableState(): DataTableState {
+  get dataTableState(): TransmartTableState {
     return this._dataTableState;
   }
 
-  set dataTableState(value: DataTableState) {
+  set dataTableState(value: TransmartTableState) {
     this._dataTableState = value;
   }
 
--- a/src/app/models/query-models/query.ts
+++ b/src/app/models/query-models/query.ts
@@ -25,6 +25,8 @@ export class Query {
   private _observationsQuery: {data: string[]};
   // The visual indicator flags the visibility of the query
   private _visible: boolean;
+  // Additional information about the query, e.g. a data table state
+  private _queryBlob: string;
 
   /*
    * Subscription feature
@@ -154,6 +156,14 @@ export class Query {
 
   set visible(value: boolean) {
     this._visible = value;
+  }
+
+  get queryBlob(): string {
+    return this._queryBlob;
+  }
+
+  set queryBlob(value: string) {
+    this._queryBlob = value;
   }
 
   get createDateInfo(): string {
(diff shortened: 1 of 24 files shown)


src/app/models/query-models/query.ts
```diff
@@ -25,6 +25,8 @@ export class Query {
   private _observationsQuery: {data: string[]};
   // The visual indicator flags the visibility of the query
   private _visible: boolean;
+  // Additional information about the query, e.g. a data table state
+  private _queryBlob: string;
```
beaugogh: why is it called queryBlob, not sth. like tableState?

beaugogh: also about the type of table state, it would be better to have it structured instead of put everything into a string. The tables states that we need to store: row dimensions, column dimensions, which are readily available in the tableService, as well as the pivot row and pivot column, which we need to think of some way to represent and store.

beaugogh: we probably need to rename patientQuery field to subjectQuery, and give it a more appropriate type, e.g. Constraint

src/app/models/table-models/data-table-state.ts
```diff
@@ -0,0 +1,14 @@
+import {Order} from "./order";
+
+export class DataTableState {
+  public rowDimensions: Array<string>;
```
beaugogh: This would be an array of Dimensions, Array<Dimension>

src/app/services/table.service.ts
```diff
@@ -1,22 +1,46 @@
...
+import {DataTableState} from "../models/table-models/data-table-state";
+import {Order} from "../models/table-models/order";
 
 @Injectable()
 export class TableService {
 
   private _candidates: Dimension[];
   private _rows: Dimension[];
   private _columns: Dimension[];
+  private _sorting : Map<string, Order>;
```
beaugogh: not sure I understand here, how can we tell which are the pivot row and column, which are needed for sending API call to the backend ?

src/app/models/query-models/query.ts
ewelinagr: it's 'queryBlob', because the purpose of this field is to store all additional information about the query, not only the tableState. For subject/observationsQuery it's different, they are stored in separate columns in the database

ewelinagr: and the query is structured, it is saved as an instance of DataTableState and parsed in the same way

src/app/models/table-models/data-table-state.ts
ewelinagr: but in back-end we store only names of dimensions, we don't need additional info, that's why saving and restoring functions are using it as an array of strings

src/app/services/table.service.ts
ewelinagr: it can be distinguished based on what was saved as a row and what was saved as a column

src/app/models/query-models/query.ts
beaugogh: I see, still, the queryBlob is a low-level design consideration that does not belong here, we need to store the criteria/constraint/query for each step separately,  and typed, subjectQuery would be of a constraint type, the observationQuery is already structured with {data: string[]}, so that is fine for now, and the tableQuery is of DataTableState type. If the backend needs the blob string, the query class provides a function that assembles all three criteria into a string, and this function will be used by the resource service.

src/app/models/table-models/data-table-state.ts
beaugogh: backend design does not influence how we design table state, the same logic as the design for the query class.

src/app/services/table.service.ts
beaugogh: what's the key of this map? is it 'rows' and 'cols'?

src/app/services/query.service.ts
```diff
@@ -592,18 +594,23 @@ export class QueryService {
   public saveQuery(queryName: string) {
```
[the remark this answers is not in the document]
ewelinagr: done by introducing a TransmartQuery model

src/app/models/query-models/query.ts
```diff
@@ -25,6 +26,8 @@ export class Query {
   private _observationsQuery: {data: string[]};
   // The visual indicator flags the visibility of the query
   private _visible: boolean;
+  // The information about the saved data table state
+  private _dataTableState: TransmartTableState;
```
beaugogh: so the models inside the constraint-models, query-models, and table-models are what we call glowing bear models, they are the models that glowing bear operates on, and they ought not to know the existence of backend-specific models such as transmart-resource models. We need to create a service that translates backend-specific models to glowing bear models

src/app/services/query.service.ts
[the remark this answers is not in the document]
beaugogh: query service will still be operating on standard glowing bear models, i.e. it does not know the existences of transmart-models, we need to create another service that translates these transmart-specific responses into the standard ones that glowing bear understands.

PR #308, opened 2018-03-20
add data-table and row models, modify dimension model, prepare mock data 

beaugogh:


src/app/services/table.service.ts
```diff
@@ -1,39 +1,173 @@
...
+  private _dataTable: DataTable;
 
   constructor() {
-    this.rows = [];
-    this.columns = [];
+    this.dataTable = new DataTable();
+    this.mockData();
+  }
+
+  mockData() {
```
beaugogh: the mocked data is only used for now to quickly create primeng table, later when there is backend connection, this function will be moved to unit test.

PR #309, opened 2018-03-21
combination constraint in patient selection needs to use subselection…

beaugogh:
… to select patients


jochemb:
@beaugogh We are seeing the problem that is solved by this on the test and dev environments. What is the reason this has not been merged?

beaugogh:
this has been solved with another PR

PR #310, opened 2018-03-22
 resource service refactor

beaugogh:


beaugogh:
@ewelinagr still need to change sth. will let you know when it's ready!

src/app/models/table-models/sort-order.ts
```diff
@@ -0,0 +1,4 @@
+export enum SortOrder {
```
ewelinag: it's duplicating table-models/order.ts model, we should remove one of them

ewelinagr: it's duplicating table-models/order.ts model, we should remove one of them

PR #312, opened 2018-03-26
 add primeng table with mock data

beaugogh:


src/app/modules/gb-data-selection-module/table-components/gb-table-dimensions/gb-table-dimensions.component.html
```diff
@@ -7,8 +7,15 @@
...
               [showTargetControls]="false"
-              [sourceStyle]="{'height':'300px'}"
-              [targetStyle]="{'height':'300px'}">
+              (onMoveToTarget)="updateTable()"
+              (onMoveToSource)="updateTable()"
+              (onMoveAllToTarget)="updateTable()"
+              (onMoveAllToSource)="updateTable()"
+              (onSourceReorder)="updateTable()"
+              (onTargetReorder)="updateTable()"
+              (drop)="onDrop()"
```
ewelinagr: wouldn't it be better to add some button to apply changes after setting a desired dimension state? The back-end call is kind of complex, so maybe it's better to limit the number of calls?

src/app/modules/gb-data-selection-module/table-components/gb-table-dimensions/gb-table-dimensions.component.ts
```diff
@@ -9,15 +9,47 @@ import {TableService} from '../../../../services/table.service';
...
+
+  updateTable() {
+    this.tableService.mockDataUpdate();
+  }
+
+  /**
+   * This function handles the drop event when the user is reordering dimensions within
+   * the same row-dimension container or the same column-dimension container
+   */
+  onDrop() {
```
ewelinagr: why is this needed? is the [dragdrop]=true option not working correctly?

src/app/modules/gb-data-selection-module/table-components/gb-table-table/gb-table-table.component.ts
```diff
@@ -0,0 +1,26 @@
...
+import {TableService} from '../../../../services/table.service';
+import {Col} from '../../../../models/table-models/col';
+import {Row} from '../../../../models/table-models/row';
+
+@Component({
+  selector: 'gb-table-table',
+  templateUrl: './gb-table-table.component.html',
+  styleUrls: ['./gb-table-table.component.css']
+})
+export class GbTableTableComponent implements OnInit {
```
ewelinagr: I am not sure if it's on purpose, but maybe better to rename it to just GbTableComponent?

beaugogh: yes I will rename it! but to differentiate it from the existing gb-table, I will name it to gb-table-view

src/app/modules/gb-data-selection-module/table-components/gb-table-dimensions/gb-table-dimensions.component.ts
beaugogh: primeng does not provide this event, when the user drags to reorder the dimensions, nothing happens! typical

ewelinagr: hmm, the demo on their website seemed working, but then probably they broke something. Ok, so please ignore my comment here

PR #313, opened 2018-03-27
 Include a table state in the runExportJob call

ewelinagr:
TMT-87


src/app/services/resource.service.ts
```diff
@@ -155,14 +155,22 @@ export class ResourceService {
...
+               dataTable?: DataTable): Observable<ExportJob> {
+    if (dataTable){
+      const transmartTableState: TransmartTableState = TransmartMapper.mapDataTable(dataTable);
+      return this.transmartResourceService
+        .runExportJob(jobId, constraint, elements, includeMeasurementDateColumns, transmartTableState);
+    } else {
+      return this.transmartResourceService
+        .runExportJob(jobId, constraint, elements, includeMeasurementDateColumns);
+    }
   }
```
beaugogh: transmartResourceService.runExportJob can accept transmartTableState as an optional parameter, we can get rid of the if/else statement

PR #315, opened 2018-03-29
fix import step 2 criteria issue: use the correct leaf nodes to compo…

beaugogh:
…se projection constraint


ewelinagr:
Maybe this fix should go to 'dev' branch, so it would be easier to rebase 'luke' branch on this?

PR #318, opened 2018-03-29
Replace mocked dimensions for step 3 with api call

ewelinagr:


src/app/services/resource.service.ts
```diff
@@ -259,4 +261,36 @@ export class ResourceService {
...
   }
 
+  /**
+   * Gets all elements from the study dimension that satisfy the constaint if given
+   * @param {Constraint} constraint
+   * @returns {Observable<string[]>}
+   */
+  getStudyNames(constraint: Constraint): Observable<string[]> {
+    return this.transmartResourceService.getStudyNames(constraint)
+      .map((elements: TransmartStudyDimensionElements[]) => {
```
beaugogh: 'TransmartStudyDimensionElements' is plural? is that correct?

beaugogh: dont we already have the resourceService.getStudies() call for this?

src/app/services/resource.service.ts
```diff
@@ -259,4 +261,36 @@ export class ResourceService {
...
+      });
+  }
+
+  /**
+   * Gets available dimensions for step 3
+   * @param studyNames
+   * @returns {Observable<string[]>}
+   */
+  getAvailableDimensions(studyNames: string[]): Observable<string[]> {
+    return this.transmartResourceService.getAvailableDimensions(studyNames)
```
beaugogh: study names can be obtained from the studies inside constraint service

src/app/services/table.service.ts
```diff
@@ -158,6 +160,25 @@ export class TableService {
     );
   }
 
+  getDimensions() {
+    const selectionConstraint = this.constraintService.generateSelectionConstraint();
+    const projectionConstraint = this.constraintService.generateProjectionConstraint();
+    let combo = new CombinationConstraint();
+    combo.addChild(selectionConstraint);
+    combo.addChild(projectionConstraint);
+    this.resourceService.getStudyNames(combo).subscribe(
```
beaugogh: the study names are already obtained by in the constructor of constraint service

src/app/services/resource.service.ts
ewelinagr: ah... no, I'll change that

ewelinagr: resourceService.getStudies() returns all the studies

src/app/services/resource.service.ts
ewelinagr: but this has be a list only of studies that are satisfying all the constraints from previous steps

src/app/services/resource.service.ts
beaugogh: then shouldn't we modify the existing getStudies call so that it takes  constraint as input and outputs the corresponding Study[] ?

ewelinagr: to get Study[] based on constraint there are 2 back-end calls needed. First `v2/dimensions/study/elements` to get just the study names based on constraints and than `studies/studyId/${studyNames}` to get Study objects. getStudies() is used only for auto-complete, constraints are not needed there, right (than it can use only one back-end call)?

beaugogh: just one more question, in future, there will be filtering option on a dimension, e.g. the user may indicate that for the study dimension, he only wants to see study-A, and study-B, but not study-C, so the glowing bear should be able to tell transmart to return a customized table based on this filtering, by sending an API call that includes the filtering study names, right? Are the study names in fact study IDs that can be used to make the API call? If they are in fact IDs, then are they really readable by users?

beaugogh:
just one more question, in future, there will be filtering option on a dimension, e.g. the user may indicate that for the study dimension, he only wants to see study-A, and study-B, but not study-C, so the glowing bear should be able to tell transmart to return a customized table based on this filtering, by sending an API call that includes the filtering study names, right? Are the study names in fact study IDs that can be used to make the API call? If they are in fact IDs, then are they really readable by users?

ewelinagr:
Yes, this is a bit confusing, but studyIds are in fact study names, so the readable ones. There are also ids of the studies, which are in fact study numbers, these are just internal database numbers. So study.id (=study_num) is not the same as study.studyId (=study_name)

ewelinagr:
back-end calls can use both study names and numbers, so this shouldn't be a problem for filtering

beaugogh:
ok, clear, thanks, I will merge it!

PR #320, opened 2018-04-05
Table refactor

beaugogh:


src/app/services/query.service.ts
```diff
@@ -170,8 +168,7 @@ export class QueryService {
     this.treeNodeCountsUpdate = appConfig.getConfig('tree-node-counts-update', true);
     this.countsRelay = false;
     this.autosaveSubjectSets = appConfig.getConfig('autosave-subject-sets', false);
-    this.exportDataView = appConfig.getConfig('export-data-view', 'default');
-    this.loadQueries();
+    // this.loadQueries();
```
ewelinagr: Why is this commented out? I don't see other place where loadQueries() is used.

beaugogh: yes, this is my mistake, will change that!

PR #322, opened 2018-04-11
Table verify

beaugogh:


src/app/services/transmart-resource/transmart-resource.service.ts
```diff
@@ -484,15 +484,38 @@ export class TransmartResourceService {
...
+      columnDimensions: colDims,
       offset: offset,
       limit: limit
     };
+    if (tableState.rowDimensionSorting) {
+      let sort = [];
+      tableState.rowDimensionSorting.forEach((val, key) => {
+        sort.push([key, val]);
+      });
+      body['rowSort'] = sort;
```
ewelinagr: Here we need to add one exception - when there is only one rowSort dimension. The table endpoint accepts either a list of dimensions, example: `['study', 'patient']` (then by default the ascending order is applied) or list of tuples, example: `[['study','asc'], ['patient', 'desc']]`, so as implemented. Different situation is only for a single dimension, then it cannot be represented as  `['study', 'desc']`, because then 'desc' would be interpreted as a dimension name an an exception would accure.
For a single dimension case this parameter should be send as `['study': 'desc']` map pair.

src/app/services/transmart-resource/transmart-resource.service.ts
```diff
@@ -484,15 +484,38 @@ export class TransmartResourceService {
...
+    }
+    if (tableState.columnDimensionSorting) {
+      let sort = [];
+      tableState.columnDimensionSorting.forEach((val, key) => {
+        sort.push([key, val]);
+      });
+      body['columnSort'] = sort;
```
ewelinagr: Same situation as described above.

src/app/services/transmart-resource/transmart-resource.service.ts
beaugogh: hmm, very bad format design for the api call, but I will change it

src/app/services/transmart-resource/transmart-resource.service.ts
beaugogh: will do

PR #324, opened 2018-04-13
PIC-SURE compatibility: authentication w/ OpenID Connect

mickmis:
- logic of OIDC is in AppModule & AppComponent
- removed authentication with OAuth2 (also the EndPointService)
- added an auto-login component that redirects to the OIDC server (known with idp-url config key)
- token is injected in headers using HTTP interceptor


beaugogh:
fix the tests first?

src/app/app.component.ts
```diff
@@ -23,13 +18,47 @@ export class AppComponent implements OnInit {
   private x_pos: number; // Stores x coordinate of the mouse pointer
   private x_gap: number; // Stores x gap (edge) between mouse and gutter
 
-  constructor(private endpointService: EndpointService,
-              private resourceService: ResourceService,
-              private transmartResourceService: TransmartResourceService,
-              private treeNodeService: TreeNodeService,
-              private constraintService: ConstraintService,
-              private queryService: QueryService,
-              private tableService: TableService) {
```
beaugogh: these services are injected here to be accessible throughout the entire application as singular instances

src/app/app.component.ts
```diff
@@ -23,13 +18,47 @@ export class AppComponent implements OnInit {
...
+  }
+
+  private onAuthorizationResultComplete(authorizationResult: AuthorizationResult) {
+    const path = JSON.parse(localStorage.getItem('redirect'));
+
+    if (authorizationResult === AuthorizationResult.authorized) {
+      this.router.navigate([path]);
+    } else {
+      this.router.navigate(['/']);
+    }
```
beaugogh: feels like there auth procedures can be put in 'endpoint service' or 'authentication service' or other names, so that the code is more modularized

src/app/app.component.ts
```diff
@@ -93,13 +122,12 @@ export class AppComponent implements OnInit {
     window.addEventListener('resize', onResize.bind(this));
   }
 
+  ngOnDestroy(): void {
+    this.oidcSecurityService.onModuleSetup.unsubscribe();
+  }
+
   logout() {
```
beaugogh: endpoint service can be the delegate or representative of oidcSecurityService that glowing bear app talks to

src/app/app.module.ts
```diff
@@ -40,25 +53,68 @@ export function initConfig(config: AppConfig) {
...
+      openIDImplicitFlowConfiguration.forbidden_route = '/';
+      openIDImplicitFlowConfiguration.unauthorized_route = '/';
+
+      // oidc well known endpoints configuration (loaded in initConfig())
+      const authWellKnownEndpoints = new AuthWellKnownEndpoints();
+      authWellKnownEndpoints.setWellKnownEndpoints(this.oidcConfigService.wellKnownEndpoints);
+
+      this.oidcSecurityService.setupModule(openIDImplicitFlowConfiguration, authWellKnownEndpoints);
+    });
+  }
```
beaugogh: this could also go into the constructor of endpoint/authentication service

src/app/utilities/HttpHeadersInterceptor.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (token !== '') {
+      return req.clone({ setHeaders: {
+          'Content-Type': 'application/json',
+          'Authorization': 'Bearer ' + this.oidcSecurityService.getToken()
+        } });
+    } else {
+      throw new Error('Token is null');
+    }
+  }
+}
```
beaugogh: nice interceptor!

src/app/app.component.ts
mickmis: that's actually a mistake, will change back

mickmis: Wait - actually is it needed to be declared in AppComponent, if it is not used in AppComponent? Isn't it enough to be declared in the subcomponents or services only when they are used?
From my understanding as long as it is declared as a provider in the root module (AppModule), the injected instances will be singular, but I might be wrong.

src/app/app.component.ts
mickmis: OK, this is going in "AuthService"

src/app/app.component.ts
mickmis: OK, this is going in "AuthService"

src/app/app.component.ts
beaugogh: maybe give a full name such as AuthenticationService, following the naming convention in glowing bear

mickmis: Will do

PR #325, opened 2018-04-16
 revise table call integration, to be continued

beaugogh:


beaugogh:
@ewelinagr not finished yet, todo: map transmart table to correct datatable format

PR #341, opened 2018-04-19
Replace Authentication Protocol: use OpenID Connect

mickmis:
- logic is in AuthenticationService, which replace EndPointService
- also includes a new AutoLoginComponent to prevent loading before login
- token is injected in headers using HTTP interceptor
- some loading calls are (temporarily) disabled for working tests


src/app/app.component.spec.ts
```diff
@@ -6,14 +6,14 @@ import {AppConfig} from './config/app.config';
 import {APP_INITIALIZER} from '@angular/core';
 import {BrowserModule} from '@angular/platform-browser';
 import {FormsModule} from '@angular/forms';
-import {HttpModule} from '@angular/http';
+import {HttpClientModule} from '@angular/common/http';
 import {BrowserAnimationsModule} from '@angular/platform-browser/animations';
 import {ResourceService} from './services/resource.service';
-import {EndpointService} from './services/endpoint.service';
+import {AuthenticationService} from './services/authentication.service';
```
beaugogh: just thought about this: do we have an idea how to deal with authorization in future? create another service called authorization.service? or manage both authentication and authorization in the same service?

src/app/app.component.ts
```diff
@@ -94,12 +94,6 @@ export class AppComponent implements OnInit {
   }
 
   logout() {
-    this.resourceService.logout()
-      .subscribe(
-        res => {
-          this.endpointService.invalidateToken();
-        }
-      );
+    this.authService.logout();
```
beaugogh: nice

src/app/config/config.dev.json
```diff
@@ -1,5 +1,10 @@
 {
-  "api-url": "http://localhost:8080",
+  "api-url": "http://localhost:8082/IRCT-CL/rest",
   "api-version": "v2",
-  "app-url": "http://localhost:4200"
+  "app-url": "http://localhost:4200",
+
+  "idp-url": "http://localhost:8081/auth/realms/master",
```
beaugogh: what is idp-url?

src/app/services/table.service.ts
```diff
@@ -19,8 +19,8 @@ export class TableService {
     this.dataTable = new DataTable();
     this.prevRowDimensions = [];
     this.prevColDimensions = [];
-    this.mockDataUpdate();
-    this.getDimensions();
+    // todo: this.mockDataUpdate();
+    // todo: this.getDimensions();
```
beaugogh: looks like an old version of the table branch, since it has been merged into dev, maybe directly rebase this PR with dev

src/app/config/config.dev.json
mickmis: IDentity Provider URL

src/app/app.component.spec.ts
mickmis: I am not sure about that. 
The thing is that while it is relatively easy to have common authentication across back end systems, authorisations are more system-specific.
Which means that within Glowing Bear handling authentication is fine, but authorisations are more tricky.
Because of that, in the future there will be a choice that will need to be made about authorisations:
- find a common ground across back end systems and enforce them in GB
- let the back end systems deal with the authorisation

IMO this will require more thinking.

src/app/auto-login.component.ts
```diff
@@ -0,0 +1,31 @@
+import { Component, OnInit, OnDestroy } from '@angular/core';
+import { OidcSecurityService } from 'angular-auth-oidc-client';
+
+@Component({
+  selector: 'gb-app-auto-login',
```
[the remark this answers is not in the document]
mickmis: OK, just committed a fix.

src/app/services/table.service.ts
mickmis: OK, will rebase on dev.

mickmis:
I performed a rebase on today's dev branch (25.04.18) and made some changes so that the authentication modification is more generic and can be used as well with tranSMART.
Okay for merge?

PR #344, opened 2018-04-23
Dev revise

beaugogh:


ewelinagr:
I'm just comparing the branches, what about https://github.com/thehyve/glowing-bear/commit/2a99a042ff03b02893ce62f497742ef7f494bc6e'? Is it needed?

beaugogh:
@ewelinagr yes, that is used by puppet to build glowing bear, since we are now using yarn, the tasks are switched to yarnTasks as well

ewelinagr:
@beaugogh ok, git showed me this change is not in dev yet, but apparently it is, great! So I checked other changes compared to the luke branch, looks like everything is in this commit! The only difference I found is https://github.com/thehyve/glowing-bear/commit/ee46b2bd746a4be96498edbf8538d3fe11824b81, (change for the gb-tree-nodes-component.ts), but I am not sure if it is needed. Other than that, I am fine with merging this commit.

PR #353, opened 2018-05-01
Update export parameters

ewelinagr:


src/app/models/export-models/export-data-type.ts
```diff
@@ -4,11 +4,13 @@ export class ExportDataType {
   private _name: string;
   private _checked: boolean;
   private _fileFormats: ExportFileFormat[];
+  private _dataView: string;
 
-  constructor(name: string, checked: boolean) {
+  constructor(name: string, checked: boolean, dataView: string) {
```
beaugogh: so 'dataView' is sth. specific to tranmart 17.1, we created this to accommodate NTR?  I would move this down to transmart resource service, as before

beaugogh: frontend models serve mainly for creating visual elements, e.g. for export-data-type, the name is for displaying the name of the data type, the checked flag is used to bind to a checkbox, fileFormats are used to display available file formats, each field has a role in the DOM, but dataView is sth. there purely for making backend API calls, which is why we move it down to the transmart.resource layer. We don't have a requirement for the user to see which data view he is exporting, right? that would be confusing anyway if the user sees 'surveyTable'

PR #354, opened 2018-05-03
Fix mapping of table dimensions

ewelinagr:
TMT-209


beaugogh:
There is still sth. strange going on, e.g. if I select 'Vital Signs' in step 2 and then go to step 3, I get concept, study and patient dimensions in that order, I then move the concept dimension to columns and move the patient dim above the study dim, then click 'update', but the ordering between study and patient is switched back to 'first study, then patient'

PR #355, opened 2018-05-14
Update data-table sort parameters parsing

ewelinagr:
TMT-220

To be merged after [PR#224](https://github.com/thehyve/transmart-core/pull/224)


ewelinagr:
@beaugogh PR#224 is merged on the back-end side, so this one can be now merged as well.

src/app/models/transmart-models/transmart-data-table.ts
```diff
@@ -1,14 +1,14 @@
 import {TransmartDimension} from './transmart-dimension';
 import {TransmartRow} from 'app/models/transmart-models/transmart-row';
-import {TransmartSorting} from './transmart-sorting';
+import {TransmartSort} from './transmart-sorting';
```
beaugogh: try to keep the naming consistent, TransmartSort - transmart-sort.ts

ewelinagr: ah sorry, I forgot to rename the file, I will update it, thanks!

PR #356, opened 2018-05-15
Fix parsing of data-table with nullable dimensions

ewelinagr:
TMT-217

### DIFF
--- a/src/app/services/transmart-resource/transmart-mapper.ts
+++ b/src/app/services/transmart-resource/transmart-mapper.ts
@@ -97,20 +97,28 @@ export class TransmartMapper {
     // get row dimensions
     transmartTable.row_dimensions.forEach((rowDim: TransmartDimension) => {
       let rowDimension = new Dimension(rowDim.name);
-      let elements = this.parseObjectToMap(rowDim.elements);
-      elements.forEach((value: Map<string, object>, key: string) => {
-        rowDimension.values.push(new DimensionValue(key, this.getDimensionMetadata(rowDim.name, value)));
-      });
+      if (rowDim.elements != null) {
+        let elements = this.parseObjectToMap(rowDim.elements);
+        elements.forEach((value: Map<string, object>, key: string) => {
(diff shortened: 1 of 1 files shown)


src/app/services/transmart-resource/transmart-mapper.ts
```diff
@@ -175,9 +183,14 @@ export class TransmartMapper {
       // get row dimensions
       transmartTable.rows[i].dimensions.forEach((inRowDim: TransmartInRowDimension) => {
         if (inRowDim.key == null) {
-          // if dimension is inline
-          newRow.addDatum(inRowDim.element[headerNameField],
-            this.getDimensionMetadata(inRowDim.dimension, inRowDim.element));
+          if (inRowDim.key == null) {
```
beaugogh: two times if (inRowDim.key == null) ?

src/app/services/transmart-resource/transmart-mapper.ts
```diff
@@ -175,9 +183,14 @@ export class TransmartMapper {
       // get row dimensions
       transmartTable.rows[i].dimensions.forEach((inRowDim: TransmartInRowDimension) => {
         if (inRowDim.key == null) {
```
beaugogh: try to use === instead of ==

ewelinagr: this was on purpose, because `==` checks both `null` and `undefined`

ewelinagr: in a current data-table json representation the dimension object in a row has name and either a key or an element. Key means that more details are described in dimension description part, while the element means that the dimension is inline - has one value (eg. start time dimension) so there is no need to add a key there, easier to just put inline value. 

So either key is undefined or element. There is also an option that both are null (eg. some observation does not have a start time or end time).

As I understand from TypeScript guides, '==' was meant to be used instead of '===null or ===undefined', but if it is less readable for you, I can change it.

PR #357, opened 2018-05-16
Fix importing and restoring queries

ewelinagr:
TMT-214,
TMT-216,
TMT-208

### DIFF
--- a/src/app/models/constraint-models/negation-constraint.ts
+++ b/src/app/models/constraint-models/negation-constraint.ts
@@ -23,29 +23,29 @@ export class NegationConstraint implements Constraint {
     return 'NegationConstraint';
   }
 
-  toQueryObjectWithSubselection(): Object {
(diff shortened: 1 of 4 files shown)


src/app/models/constraint-models/negation-constraint.ts
```diff
@@ -23,29 +23,29 @@ export class NegationConstraint implements Constraint {
     return 'NegationConstraint';
   }
 
-  toQueryObjectWithSubselection(): Object {
+  toQueryObjectWithSubselection(full?: boolean): Object {
```
beaugogh: is this flag (full) needed for negation constraint?

ewelinagr: yes, because previously when you added concept constraint to exclusion criteria, only the path was saved in the query (if you did this for inclusion the full data for concept was added).  Because of this it was impossible to restore the saved query with excluded concept constraint (TMT-216)

PR #359, opened 2018-05-17
Change naming of exported files, fix export

ewelinagr:
TMT-213


src/app/modules/gb-export-module/gb-export.component.ts
```diff
@@ -16,7 +17,8 @@ export class GbExportComponent implements OnInit {
   constructor(private appConfig: AppConfig,
               private exportService: ExportService,
               public resourceService: ResourceService,
-              private timer: SimpleTimer) {
+              private timer: SimpleTimer,
+              private messageService: MessageService) {
```
beaugogh: the checks are still failing, messageService should be public

PR #363, opened 2018-05-22
remove end time dimension constraint

beaugogh:


ewelinagr:
@beaugogh why do you want to remove 'end time' dimension? Is it still something failing when you use it?

ewelinagr:
Sorry, I checked it wrong, yes, of course it should be removed

PR #367, opened 2018-05-28
Replace Authentication Protocol: use OpenID Connect (merge into dev)

mickmis:
Replace OAuth2 by OpenID Connect.

* logic is in AuthenticationService, which replace EndPointService
* also includes a new AutoLoginComponent to prevent loading before login
* token is injected in headers using HTTP interceptor


src/app/config/config.default.json
```diff
@@ -8,5 +8,8 @@
   "show-observation-counts": true,
   "instant-counts-update-1": false,
   "instant-counts-update-2": false,
-  "instant-counts-update-3": false
+  "instant-counts-update-3": false,
+  "oidc-server-url": "https://keycloak.thehyve.net/auth/realms/master",
+  "oidc-scopes": "openid",
+  "oidc-client-id": "someid"
```
beaugogh: add these configs to config.default.json as well

PR #370, opened 2018-06-04
Testing gijs

gijskant:


beaugogh:
it's better to wait for the refactoring of glowingbear's workflow regarding subselection, after which, the testing should become clearer.

PR #371, opened 2018-06-04
Enable static code analysis (tslint) on Travis.

gijskant:


tslint.json
```diff
@@ -108,9 +107,6 @@
     "use-life-cycle-interface": true,
     "use-pipe-transform-interface": true,
     "component-class-suffix": true,
-    "directive-class-suffix": true,
-    "no-access-missing-member": true,
-    "templates-use-public": true,
-    "invoke-injectable": true
```
beaugogh: why disable these changes?

build.gradle
```diff
@@ -54,6 +54,13 @@ task yarnPackage(type: YarnTask, dependsOn: 'yarn') {
 }
 distTar.dependsOn(yarnPackage)
 
+task yarnLint(type: YarnTask, dependsOn: 'yarn') {
```
beaugogh: is this task run somewhere? .travis.yam?

tslint.json
gijskant: Because they are not supported by tslint, causing warnings.

build.gradle
gijskant: Check now depends on yarnLint. So, Travis will run those indeed.

PR #378, opened 2018-06-14
Support multiple authentication methods

gijskant:
- Add authentication flows for OAuth2 and OpenID Connect (OIDC).
- Add main component that only activates after authentication.

Includes #367.


gijskant:
@beaugogh A small change in the Transmart configuration is required for this: in `application.groovy` the list of supported grant types should be:
`authorizedGrantTypes: ['implicit', 'password', 'authorization_code', 'refresh_token'],`.

gijskant:
@beaugogh The `dev` branch has been merged in again.

src/app/services/transmart-services/transmart-resource.service.ts
```diff
@@ -369,19 +305,9 @@ export class TransmartResourceService {
...
-    const options = {
-      headers: new HttpHeaders({
-        'Content-Type': 'application/json',
-        'Authorization': `Bearer ${endpoint.accessToken}`
-      })
-    };
-    return this.http.get(url, {...options, responseType: 'blob'})
-      .catch(TransmartResourceService.handleError.bind(this));
+    let url = `${this.endpointUrl}/export/${jobId}/download`;
+    return this.http.get(url, {responseType: 'blob'})
```
mickmis: HTTP header "'Content-Type', 'application/zip'" is not added anymore, is it intentional?

gijskant: I wasn't aware of this change, but it seems right: the content of what we are sending to the backend is not `zip`, and that is what this header was telling.

PR #381, opened 2018-06-20
use multi-select for categorical concept values

beaugogh:


beaugogh:
* event emitters work only for parent-child component communication, not component-service communication
* remove button is intended for constraint, as it is also used by other child constraints other than combination constraints
* the other aspects are refactored

gijskant:
Great improvements!

PR #383, opened 2018-06-25
testing

beaugogh:


src/app/services/cross-table.service.spec.ts
```diff
@@ -198,134 +198,4 @@ describe('CrossTableService', () => {
     expect(spySet).toHaveBeenCalled();
   });
 
-  it('should correctly create cross constraints for a concept', () => {
```
gijskant: @beaugogh Why did you delete these tests?

src/app/services/tree-node.service.ts
```diff
@@ -185,17 +283,27 @@ export class TreeNodeService {
...
-    const fullName = treeNode['fullName'];
-    let head = fullName.substring(0, fullName.length - tail.length);
-    concept.label = treeNode['name'] + ' (' + head + ')';
-    concept.path = treeNode['conceptPath'];
-    concept.type = <ConceptType> treeNode['type'];
-    concept.code = treeNode['conceptCode'];
-    concept.fullName = treeNode['fullName'];
-    concept.name = treeNode['name'];
-    return concept;
+    if (treeNode['name'] &&
```
gijskant: You can use `treeNode.name` instead.

src/app/services/cross-table.service.spec.ts
gijskant: Covered by `cross-table.integration.spec`, right?

beaugogh: yes, should be covered by the integration tests, also, for some reason, the tests failed at these tests frequently, when I removed them, the tests passed

src/app/services/tree-node.service.ts
beaugogh: yes, I am actually thinking of refactoring the tree node type, it is something that primeNg provides, which turns out to be more and more ill-fitted for glowing bear, but not priority now

PR #385, opened 2018-06-27
Data table tests

gijskant:
Refactor data table functionality; add tests.


src/app/services/data-table.service.ts
```diff
@@ -5,11 +5,14 @@ import {Row} from '../models/table-models/row';
 import {ResourceService} from './resource.service';
 import {Col} from '../models/table-models/col';
 import {ConstraintService} from './constraint.service';
-import {HeaderRow} from '../models/table-models/header-row';
 
 @Injectable()
 export class DataTableService {
 
+  static sortableDimensions = [
```
beaugogh: transmart-specific, move it to the transmart-services or transmart-utilities

beaugogh: and it seems that these dimensions are not used anywhere?

src/app/services/resource.service.ts
```diff
@@ -40,7 +40,7 @@ export class ResourceService {
    * @returns {Observable<Study[]>}
    */
   getStudies(): Observable<Study[]> {
-    return this.transmartResourceService.getStudies();
+    return Observable.fromPromise(this.transmartStudiesService.studies);
```
beaugogh: I would cache studies in transmart resource service directly

src/app/services/resource.service.ts
```diff
@@ -287,33 +290,44 @@ export class ResourceService {
...
     let offset = dataTable.offset;
     let limit = dataTable.limit;
 
+    // Fetch dimensions for the data matching the constraint
     return this.getDimensions(dataTable.constraint)
       .switchMap((transmartStudyDimensions: TransmartStudyDimensions) => {
+        // Merge available dimensions and current table state
         let tableState: TransmartTableState =
-          TransmartMapper.mapStudyDimensionsToTableState(transmartStudyDimensions, dataTable);
+          TransmartDataTableMapper.mapStudyDimensionsToTableState(transmartStudyDimensions, dataTable);
```
beaugogh: still doubting the motivation behind separating mappers into even more mappers, if we worry about the transmart mapper getting large, we don't have to be. the mapper functions are just stand-alone, static functions, distinguishing them with comment lines would be enough. Ideally would prefer to keep all the mapper function in one place.

src/app/services/resource.service.ts
beaugogh: also, ideally glowing bear knows one backend-mapper and one backend.resource.service, we should hide backend-specific details from glowing bear as much as possible

beaugogh:
Just a general remark: when we add transmart-specific code, try to isolate such code from the 'glowing bear' code as much as possible, the principle that we try to follow is -- "glowing bear is ignorant about how transmart runs"

src/app/services/resource.service.ts
gijskant: That is not a proper way to organise your code.

beaugogh: or find alternative ways to hide these other mappers from resource service

gijskant:
@beaugogh Okay now?

beaugogh:
well, still a bit hesitant about the mapper thingy, but will merge

PR #388, opened 2018-07-02
Fix query

beaugogh:


src/app/services/data-table.service.ts
```diff
@@ -117,7 +123,9 @@ export class DataTableService {
   }
 
   set dataTable(value: DataTable) {
-    this._dataTable = value;
+    if (value instanceof DataTable) {
```
gijskant: In what situations is something passed that is not an instance of `DataTable`?

beaugogh: seems to be related to zone.js regressions, but not entirely sure, no point in the our code that assigns values other than DataTable, but while chaining promises, boolean values somehow got assigned to data table

src/app/services/query.service.ts
```diff
@@ -445,72 +474,94 @@ export class QueryService {
...
+        this.isLoadingSubjectCount_2 = true;
+        this.isLoadingObservationCount_2 = true;
+
+        this.query = null; // clear query
+        // update the subject count and observation count in the 2nd step
+        const constraint_1_2: Constraint = this.constraintService.constraint_1_2();
+        this.resourceService.getCounts(constraint_1_2)
+          .subscribe(
+            (countResponse) => {
+              const index = this.queueOfCalls_2.indexOf(timeStamp.getMilliseconds());
```
gijskant: Is this queue of calls really needed?

beaugogh: there is still some parallel calls, not every call is chained

PR #390, opened 2018-07-04
table refactoring + testing

beaugogh:


src/app/models/constraint-models/combination-constraint.ts
```diff
@@ -29,8 +30,13 @@ export class CombinationConstraint extends Constraint {
 
   addChild(constraint: Constraint) {
     if (!(constraint.className === 'CombinationConstraint'
-        && (<CombinationConstraint>constraint).isRoot)) {
-      constraint.parent = this;
+      && !(<CombinationConstraint>constraint).isRoot)) {
+      // to enforce polymorphism, otherwise child set method is not called
+      if (constraint.className === 'PedigreeConstraint') {
+        (<PedigreeConstraint>constraint).parentConstraint = this;
```
ewelinagr: Why is this required only for the PedigreeConstraint?

PR #391, opened 2018-07-09
Add updating of subjectSets before running the export job

ewelinagr:
Description of the issue: [TMT-265](https://jira.thehyve.nl/browse/TMT-265)
When autosave-subject-sets option was set to true, exported data was not up to date with a selection from step1.

The export behavior for autosave-subject-sets mode is changed to first check if step 1 subject set is up to date with criteria. If it is not, a new subject set is created and then based on it the export is created.

### DIFF
--- a/src/app/services/export.service.ts
+++ b/src/app/services/export.service.ts
@@ -60,23 +60,13 @@ export class ExportService {
               summary = 'Export job "' + name + '" is created.';
               MessageHelper.alert('success', summary);
               this.exportJobName = '';
-              if (this.constraintService.subjectSetConstraint) {
-                this.runExportJobWithSubjectSetConstraint(newJob)
-                  .then(() => {
-                    resolve(true);
-                  })
-                  .catch(err => {
-                    reject(err)
-                  });
-              } else {
-                this.runExportJob(newJob)
-                  .then(() => {
-                    resolve(true);
-                  })
-                  .catch(err => {
-                    reject(err)
-                  });
-              }
+              this.runExportJob(newJob)
+                .then(() => {
+                  resolve(true);
+                })
+                .catch(err => {
+                  reject(err)
+                });
             },
             (err: HttpErrorResponse) => {
               ErrorHelper.handleError(err);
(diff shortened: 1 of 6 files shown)


src/app/services/export.service.ts
```diff
@@ -58,13 +58,35 @@ export class ExportService {
             summary = 'Export job "' + name + '" is created.';
             MessageHelper.alert('success', summary);
             this.exportJobName = '';
+            if(this.constraintService.subjectSetConstraint) {
+              this.runExportWitSubjectSetConstraint(newJob);
```
beaugogh: typo?

beaugogh:
in constraint service, there is constraint_1() method to get the constraint in step 1, if there is already subject set constraint, this method will return the subject set constraint, so no need to modify runExport function?

ewelinagr:
this works for a general case, but when autosave-subject-set=true it uses constraints based on subject_id. Currently, to have a new subject_id, you have to click an update button (update counts), which saves a new patient set and returns back the new id from transmart. So if we want to base export on the current selection without updating counts, this modified runExport is needed.

beaugogh:
@ewelinagr rebase this one?

ewelinagr:
@beaugogh it is rebased now, but I am adding a test for this case now, should be ready soon

ewelinagr:
@beaugogh can you have a look once again?

src/app/services/transmart-services/transmart-resource.service.ts
```diff
@@ -58,6 +58,11 @@ export class TransmartResourceService {
   private _exclusionCounts: TransmartCountItem;
   private _studyConceptCountObject: object;
 
+  /**
+   * Constraints that were used to create a recently saved subject set
+   */
+  public _savedSubjectSetConstraints: Constraint;
```
beaugogh: plural?

beaugogh: also, name it sth. explicit, like '_subjectSetConstraintForExport'

src/app/services/transmart-services/transmart-resource.service.ts
```diff
@@ -279,6 +284,7 @@ export class TransmartResourceService {
       if (this.autosaveSubjectSets) {
         this.savePatientSet('temp', constraint)
           .subscribe((subjectSet: SubjectSet) => {
+            this._savedSubjectSetConstraints = constraint;
```
beaugogh: use getters instead of the field itself

src/app/services/transmart-services/transmart-resource.service.ts
```diff
@@ -517,6 +523,47 @@ export class TransmartResourceService {
...
+    return new Promise((resolve, reject) => {
+      if (this.autosaveSubjectSets
+        && this._savedSubjectSetConstraints
+        && constraint === this._savedSubjectSetConstraints) {
+        this.runExportJob(jobId, this.subjectSetConstraint, elements, tableState)
+          .subscribe((exportJob: ExportJob) => {
+            resolve(exportJob);
+          }, (error: any) => {
+            ErrorHelper.handleError(error);
+            console.error(`Error retrieving studies: ${error}`);
```
beaugogh: error retrieving studies? which studies

src/app/services/transmart-services/transmart-resource.service.ts
beaugogh: also, updateInclusionExclusionCounts is only going to be invoked with the user clicks the update_1 button, I thought we try to save subject set even though the user has not clicked the update button

ewelinagr: that was the first idea, but as we discussed yesterday, saving subject set can take some time, when there is a lot of data and might not be a good performance improvement to save it before the export to use patient set constraint instead of direct constraint from step1_2. This is something I am not sure about, so maybe we should discuss it with others.

beaugogh:
closing this one in favor of an alternative solution

PR #392, opened 2018-07-11
query service refactor

beaugogh:


src/app/config/config.dev.json
```diff
@@ -8,6 +8,7 @@
   "instant-counts-update-1": false,
   "instant-counts-update-2": false,
   "instant-counts-update-3": false,
+  "include-data-table": true,
   "authentication-service-type": "transmart",
   "oidc-server-url": "http://localhost:8081/auth/realms/master",
```
ewelinagr: I know it's not a change from this PR, but I have not noticed this in the previous one, duplicated `oidc-server-url` config and also the first one is invalid, missing `/protocol/openid-connect` part of the url.

ewelinagr:
Maybe also some tests for summary counts would be useful,

src/app/config/config.dev.json
beaugogh: indeed, I did not notice this one. you mean the dev.config should be sth. like this?

```javascript
{
  "api-url": "http://localhost:8080",
  "api-version": "v2",
  "app-url": "http://localhost:4200",
  "autosave-subject-sets": false,
  "export-data-view": "default",
  "show-observation-counts": true,
  "instant-counts-update-1": false,
  "instant-counts-update-2": false,
  "instant-counts-update-3": false,
  "include-data-table": true,
  "authentication-service-type": "transmart",
  "oidc-server-url": "http://localhost:8081/auth/realms/master",
  "oidc-client-id": "glowingbear-js"
}
```

beaugogh: what is the correct url for oidc-server-url?

ewelinagr: "oidc-server-url": "http://localhost:8080/auth/realms/master/protocol/openid-connect"

ewelinagr:
Just a small note regarding the summary counts. I think they should be change to `...` each time the button in step 2 gets `dirty`, not only after clicking the update. It would be more clear to the user I guess.

beaugogh:
@ewelinagr which it does, right?

src/app/services/resource.service.ts
```diff
@@ -63,27 +80,55 @@ export class ResourceService {
...
+              this.selectedStudyConceptCountMap =
+                TransmartMapper.mapStudyConceptCountObject(this.transmartResourceService.studyConceptCountObject);
+              resolve(true);
+            })
+            .catch(err => {
+              reject(err);
+            });
+          break;
+        }
+      }
```
ewelinagr: this `switch case Transmart:  transmartResource...; default:  transmartResource...` way requires a lot of code duplications. Maybe this could be replaced by something like endpointResource interface?

beaugogh: what do you mean?

PR #393, opened 2018-07-17
add more tests to resource service

beaugogh:


ewelinagr:
Is this bug related to some TMT ticket?

PR #395, opened 2018-07-19
Access level integration + export workflow adjustment

beaugogh:


beaugogh:
To create an export job requires at least one data type and one file type selected, which necessitates data-selection update (update1 and/or update2). In other words, to export, one must update! We make the update automatically happen when the user switches to the export tab.

ewelinagr:
To get file type and data type not the full update is needed, the counts are not required there, just the calls for data and file types. But indeed making the update would be the easiest solution, I'm not sure about the performance though.

beaugogh:
A full update is needed because export update needs the most recent constraint is step 2 as well

PR #397, opened 2018-07-23
various fixes

beaugogh:
… mode


src/app/utilities/transmart-utilities/transmart-constraint-serialiser.ts
```diff
@@ -221,6 +221,15 @@ export class TransmartConstraintSerialiser extends AbstractConstraintVisitor<obj
           };
         }
       }
+    } else if (constraint.mark === ConstraintMark.SUBJECT) {
```
ewelinagr: I do not want to conflict with your changes, so if you on this, could you also add a small update to line 177 (`===` instead of a single `=`)

beaugogh: no prob

PR #402, opened 2018-07-27
Hide data table for restricted user; disable trial visit if not present.

gijskant:


gijskant:
@ewelinagr Thanks for pointing me to the cause of the issue! Disabling the redirect to the login page in the tests resolved the issue.

src/app/modules/gb-data-selection-module/constraint-components/gb-concept-constraint/gb-concept-constraint.component.ts
```diff
@@ -250,9 +254,11 @@ export class GbConceptConstraintComponent extends GbConstraintComponent implemen
   }
 
   set selectedConcept(value: Concept) {
-    (<ConceptConstraint>this.constraint).concept = value;
-    this.initializeConstraints();
-    this.update();
+    if (value instanceof Concept) {
```
ewelinagr: is it needed when the type of the `value` parameter is Concept?

gijskant: Not having this caused issues, I think with the autocomplete. So, probably the type should be `any` then.

gijskant:
```ERROR in src/app/modules/gb-data-selection-module/constraint-components/gb-concept-constraint/gb-concept-constraint.component.ts(252,7): error TS2380: 'get' and 'set' accessor must have the same type.
src/app/modules/gb-data-selection-module/constraint-components/gb-concept-constraint/gb-concept-constraint.component.ts(256,7): error TS2380: 'get' and 'set' accessor must have the same type.```

Comment (ewelinagr) [reply src/app/modules/gb-data-selection-module/constraint-components/gb-concept-constraint/gb-concept-constraint.component.ts]: ah ok, I see

File: src/app/services/authentication/authentication.service.ts [hunk 2]
```diff
@@ -22,12 +23,82 @@ export class AuthenticationService implements OnDestroy {
   private authenticationMethod: AuthenticationMethod;
   private _accessLevel: AccessLevel = AccessLevel.Restricted;
 
+  static getAccessLevelFromRoles(roles: string[]): AccessLevel {
```
ewelinagr: I am not sure about this part, it seems that no matter what roles I choose and which user, I always have the `Full` access. What should I do to make it `Restricted`?

gijskant: If you have only `SUMMARY` or `COUNTS_WITH_THRESHOLD` access to all studies that you have access to.

gijskant: (see tests). But the code has become a bit messy, because the test script also wants to support es2015, and then you cannot use `some`, `every` or `for let roles of rolesPerStudy.values()`.

gijskant: Basically: add role `<STUDY_ID>|SUMMARY` to your user in Keycloak and make sure you're not admin.

PR #403, opened 2018-07-30
Prevent redirect when dropping an item in the cross table drop zone.

gijskant:
Fixes [TMT-361](https://jira.thehyve.nl/browse/TMT-361).
The issue only appeared in Firefox, not in Chrome.


gijskant:
I added it there now.

PR #407, opened 2018-08-07
Upgrade

beaugogh:


ewelinagr:
There are still some usages of `Observable.of` in the code, but in tests only (e.g. resource.service.spec.ts). Should they be replaced too?

beaugogh:
good point, I will update them

PR #408, opened 2018-08-09
Revert previous update flow for export

ewelinagr:
There was a change introduced in [this commit](https://github.com/thehyve/glowing-bear/commit/006c3dc61d5135f5a4667e9220aa366dc220f57e) to get rid of the error message about no export access, when clicking Step 2 update (it should not be showed there). But apparently it introduced to many different issues, so better to revert it for now and fix the error messages later. 

Fixes [TMT-386](https://jira.thehyve.nl/browse/TMT-386)

### DIFF
--- a/src/app/services/query.service.ts
+++ b/src/app/services/query.service.ts
@@ -236,6 +236,7 @@ export class QueryService {
   public update_1(initialUpdate?: boolean): Promise<any> {
     return new Promise((resolve, reject) => {
       // update export flags
+      this.exportService.isLoadingExportDataTypes = true;
       this.isUpdating_1 = true;
       // set the flags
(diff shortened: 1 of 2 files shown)


src/app/services/query.service.ts
```diff
@@ -338,6 +339,8 @@ export class QueryService {
               this.isUpdating_2 = false;
               this.isDirty_2 = false;
               this.isDirty_3 = true;
+              // update the export variables
+              this.exportService.updateExports();
```
beaugogh: previously the updateExports seems not to be here as well, should be in update_2?

beaugogh: ah, sorry, misread

PR #443, opened 2018-09-04
Hide observation counts in the tree if specified in configuration.

gijskant:
Not sure where to put the defaults for config values. Now I moved the property `showObservationCounts` from the query service to the tree node service, but ideally there is a separate injectable object that contains such properties and has the default values.

### DIFF
--- a/src/app/config/app.config.ts
+++ b/src/app/config/app.config.ts
@@ -47,7 +47,7 @@ export class AppConfig {
    */
   public getConfig(key: any) {
     let value = this.config[key];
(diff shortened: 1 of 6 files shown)


beaugogh:
why do we  need to move this flag out of the config?

beaugogh:
I agree that we need some application defaults, but showObservationCounts still needs to be read from config rather than directly using a default

gijskant:
@beaugogh The PR has been updated with the latest changes in dev.

PR #445, opened 2018-09-04
Store version in version.json file, read by both app and gradle.

gijskant:
With these changes, version updates have to be put in only two places:
- `src/app/config/version.json` (used by the application and by gradle)
- `package.json`

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,14 @@ e.g. [Apache] or [nginx].
 
 
 
+### How to create a release
+
+For creating a new release, increase the version in [version.json](src/app/config/version.json)
+and [package.json](package.json). Make sure the `publishing.repositories.maven.url` property
+in [build.gradle](build.gradle) is set to a release repository.
+
+
+
 ### Configuration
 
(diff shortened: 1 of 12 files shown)


beaugogh:
there is a ticket corresponding to this issue, TMT-56, plz follow the comments there

beaugogh:
as what we have talked about, the puppet does not need to overwrite everything in the config.default.json. In principle, puppet shouldn't overwrite anything in the config, different deployment of glowingbear will have different config.default.json's, which is the only place where glowingbear-related config is set. A secondary approach would be puppet overwriting only the part it needs for individual glowingbear deployment, leaving the rest of the config options untouched.

beaugogh:
Using as few config (or manual input) files as possible would be recommended in our future development

gijskant:
Having as few as possible configuration files is one thing, I my view it is also good to have sensible defaults and only having to configure specific config values in the config file.

gijskant:
Close this in favour of #446, which stores the version in `config.default.json` instead of in a separate `version.json` file.

PR #447, opened 2018-09-05
refactor config

beaugogh:


src/app/config/app.config.ts
```diff
@@ -14,7 +14,24 @@ import {ErrorHelper} from '../utilities/error-helper';
 @Injectable()
 export class AppConfig {
 
+  public static DEFAULT_API_URL = 'https://transmart-dev.thehyve.net';
+  public static DEFAULT_API_VERSION = 'v2';
+  public static DEFAULT_APP_URL = 'http://localhost:4200';
+  public static DEFAULT_APP_VERSION = '0.0.1-SNAPSHOT';
```
gijskant: Maybe the default app version should be something like 'no version specified'

src/app/config/app.config.ts
```diff
@@ -14,7 +14,24 @@ import {ErrorHelper} from '../utilities/error-helper';
 @Injectable()
 export class AppConfig {
 
+  public static DEFAULT_API_URL = 'https://transmart-dev.thehyve.net';
```
gijskant: Should this have a default? I would see this as always environment specific.

src/app/config/app.config.ts
beaugogh: good idea, will change

src/app/config/app.config.ts
beaugogh: yes, we should always specify this in config.default.json, on an off-chance that this url is not there, would be nice to have sth. to fall back on?

PR #449, opened 2018-09-07
Fix the selection of the surveyTable view for TSV export.

gijskant:


src/app/utilities/transmart-utilities/transmart-mapper.ts
```diff
@@ -136,11 +136,7 @@ export class TransmartMapper {
             let el = new TransmartExportElement();
             el.dataType = dataType.name;
             el.format = fileFormat.name;
-            if (fileFormat.name === 'TSV' && dataType.name === 'clinical') {
-              el.dataView = 'dataTable';
-            } else {
-              el.dataView = defaultDataView;
-            }
+            el.dataView = defaultDataView;
```
beaugogh: why the change here?

gijskant: If the dataView is set to 'surveyTable', the TSV export would still choose 'dataTable' instead. The idea was that this setting determines what dataView is selected for the exports.

PR #455, opened 2018-10-08
 Add custom export job handling

ewelinagr:
- add TransmartExternalJobResourceService to handle calls to external tools, like [transmart-packer](https://github.com/thehyve/transmart-packer)  
- use the external tool to create custom export jobs

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ Supported properties in the `config.*.json` files:
 | `api-version`             | `v2`      | TranSMART API version. Only `v2` is supported. |
 | `app-url`                 |           | URL where the Glowing Bear is accessible for the user.|
 | `autosave-subject-sets`   | `false`   | Persist subject selection as subject set automatically. |
-| `export-data-view`        | `default` | Shape of the export (`default`, `surveyTable`). |
+| `export-data-view`        | `default` | Shape of the export (`default`, `surveyTable`, `customFormat`. |
 | `show-observation-counts` | `true`    | |
 | `instant-counts-update-1` | `false`   | |
 | `instant-counts-update-2` | `false`   | |
@@ -143,6 +143,8 @@ Supported properties in the `config.*.json` files:
 | `authentication-service-type` | `oidc`  | Authentication service type (`oidc`, `transmart`) |
 | `oidc-server-url`         |           | E.g., `https://keycloak.example.com/auth/realms/{realm}/protocol/openid-connect` |
 | `oidc-client-id`          | `glowingbear-js` | |
+| `export-service-url`      |           | URL of the external tool that handles the export, when the `export-data-view` is set to `customFormat`. |
+| `custom-export-job-name`  |           | Name of the export job that will be triggered in the external tool, when the `export-data-view` is set to `customFormat`. |
 
 
 ## License
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ Supported properties in the `config.*.json` files:
 | `api-version`             | `v2`      | TranSMART API version. Only `v2` is supported. |
 | `app-url`                 |           | URL where the Glowing Bear is accessible for the user.|
 | `autosave-subject-sets`   | `false`   | Persist subject selection as subject set automatically. |
-| `export-data-view`        | `default` | Shape of the export (`default`, `surveyTable`, `customFormat`. |
+| `export-data-view`        | `default` | Shape of the export (`default`, `surveyTable`, `customFormat`). |
 | `show-observation-counts` | `true`    | |
 | `instant-counts-update-1` | `false`   | |
 | `instant-counts-update-2` | `false`   | |
--- a/README.md
+++ b/README.md
@@ -135,7 +135,6 @@ Supported properties in the `config.*.json` files:
 | `api-version`             | `v2`      | TranSMART API version. Only `v2` is supported. |
 | `app-url`                 |           | URL where the Glowing Bear is accessible for the user.|
 | `autosave-subject-sets`   | `false`   | Persist subject selection as subject set automatically. |
-| `export-data-view`        | `default` | Shape of the export (`default`, `surveyTable`, `customFormat`). |
 | `show-observation-counts` | `true`    | |
 | `instant-counts-update-1` | `false`   | |
 | `instant-counts-update-2` | `false`   | |
@@ -143,8 +142,10 @@ Supported properties in the `config.*.json` files:
 | `authentication-service-type` | `oidc`  | Authentication service type (`oidc`, `transmart`) |
 | `oidc-server-url`         |           | E.g., `https://keycloak.example.com/auth/realms/{realm}/protocol/openid-connect` |
 | `oidc-client-id`          | `glowingbear-js` | |
-| `export-service-url`      |           | URL of the external tool that handles the export, when the `export-data-view` is set to `customFormat`. |
-| `custom-export-job-name`  |           | Name of the export job that will be triggered in the external tool, when the `export-data-view` is set to `customFormat`. |
+| `use-external-job`        | `false`   | Use an external tool that handles the export. Otherwise the dataView will be used. |
+| `export-data-view`        | `default` | Shape of the export (`default`, `surveyTable`).
+| `export-service-url`      |           | URL of the external tool that handles the export, if the `use-external-job` is set to `true`. |
+| `custom-export-job-name`  |           | Name of the export job that will be triggered in the external tool, if the `use-external-job` is set to `true`. |
 
 
 ## License
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -46,6 +46,8 @@ import {GrowlModule} from 'primeng/growl';
 import {GbMainModule} from './modules/gb-main-module/gb-main.module';
 import {MessageHelper} from './utilities/message-helper';
 import {of as observableOf} from 'rxjs';
+import {TransmartExternalJobResourceService} from "./services/transmart-services/transmart-external-job-resource.service";
+import {TransmartExternalJobResourceServiceMock} from "./services/mocks/transmart-external-job-resource.service.mock";
 
 export function initConfig(config: AppConfig) {
   return () => config.load();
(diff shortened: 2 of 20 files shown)


src/app/services/resource.service.ts
```diff
@@ -282,12 +286,16 @@ export class ResourceService {
   getExportDataTypes(constraint: Constraint): Observable<ExportDataType[]> {
     switch (this.endpointMode) {
       case EndpointMode.TRANSMART: {
-        return this.transmartResourceService.getExportFileFormats().pipe(
-          switchMap(fileFormatNames => {
-            return this.transmartResourceService.getExportDataFormats(constraint)
-          }, (fileFormatNames, dataFormatNames) => {
-            return TransmartMapper.mapTransmartExportFormats(fileFormatNames, dataFormatNames);
-          }));
+        if (this.transmartResourceService.exportDataView === 'customFormat') {
```
gijskant: Why ask this property from the `transmartResourceService` and not directly from the config? The config actually tells you if you should use the `transmartResourceService` for exports or not.

README.md
```diff
@@ -135,14 +135,16 @@ Supported properties in the `config.*.json` files:
 | `api-version`             | `v2`      | TranSMART API version. Only `v2` is supported. |
 | `app-url`                 |           | URL where the Glowing Bear is accessible for the user.|
 | `autosave-subject-sets`   | `false`   | Persist subject selection as subject set automatically. |
-| `export-data-view`        | `default` | Shape of the export (`default`, `surveyTable`). |
+| `export-data-view`        | `default` | Shape of the export (`default`, `surveyTable`, `customFormat`). |
```
gijskant: The `customFormat` 'dataView' is now used to signal that the 'dataView' is not being used. Perhaps add a boolean flag somewhere else, or autodetect that if the `export-service-url` is set, the external service is used and 'dataView' can be ignored (but custom-export-job-name property is used).

src/app/services/resource.service.ts
```diff
@@ -302,7 +310,14 @@ export class ResourceService {
   getExportJobs(): Observable<any[]> {
     switch (this.endpointMode) {
       case EndpointMode.TRANSMART: {
-        return this.transmartResourceService.getExportJobs();
+        if (this.transmartResourceService.exportDataView === 'customFormat') {
```
gijskant: It seems that a check like `if (this.config['use-external-job'])` would more descriptive than this condition.

src/app/services/transmart-services/transmart-external-job-resource.service.ts
```diff
@@ -0,0 +1,260 @@
...
+  // ------------------------------------- export job calls -------------------------------------
+
+  /**
+   * Make a post http request
+   * @param urlPart - the part used in baseUrl/urlPart
+   * @param body
+   * @param responseField
+   * @returns {Observable<any | any>}
+   */
+  private postCall(urlPart, body, responseField) {
```
gijskant: Such code duplication can be prevented by moving these functions to a separate helper class.

gijskant: Or have a common superclass.

src/app/services/transmart-services/transmart-external-job-resource.service.ts
```diff
@@ -0,0 +1,260 @@
...
+   *
+   * @param {string} jobName
+   * @param {Constraint} constraint
+   * @returns {Observable<ExportJob>}
+   */
+  runJob(jobName: string, constraint: Constraint): Observable<TransmartExternalJob> {
+    const urlPart = `jobs/create`;
+    const responseField = 'job';
+    let targetConstraint = constraint;
+    if (this.autosaveSubjectSets &&
```
gijskant: I don't really understand what happens here, but whatever it is, it does not seem like something that should be the responsibility of this service.

gijskant: First step would be to create a separate function that transforms the constraint in one where the subject level part is replaced with the subject set constraint. And then decide which service is responsible... the transmart resource service, or, perhaps a separate service: maybe a superclass for both?

src/app/services/transmart-services/transmart-external-job-resource.service.ts
```diff
@@ -0,0 +1,260 @@
...
+   */
+  createExportJob(name: string): Observable<ExportJob> {
+    let newExportJob = new ExportJob();
+    newExportJob.id = name;
+    newExportJob.jobName = name;
+    return observableOf(newExportJob);
+  }
+
+  /**
+   * Returns mocked file formats,
```
gijskant: Why 'mocked', these are the actual supported file formats, right?

src/app/models/transmart-models/transmart-external-job.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+export class TransmartExternalJob {
+
+  // The unique job id
+  public task_id: string;
```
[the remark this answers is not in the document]
ewelinagr: yes, but it's the mapper for transmart-packer response, which uses the snake_case, more python-style way I guess. Not sure what should be the best solution in this case, changing the packer response?

src/app/services/api-http-interceptor.service.ts
```diff
@@ -36,7 +36,8 @@ export class ApiHttpInterceptor implements HttpInterceptor {
 
     // skip if request is for config, or if not for API
     if (  req.url.includes(AppConfig.path) ||
-          !req.url.includes(this.appConfig.getConfig('api-url')) ||
+          (!req.url.includes(this.appConfig.getConfig('api-url')) &&
+          !req.url.includes(this.appConfig.getConfig('export-service-url'))) ||
```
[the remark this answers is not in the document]
ewelinagr: there is a default one, so currently it won't be empty, but you're right, this condition require some modification. Probably also the default 'export-service-url' should be removed

src/app/services/resource.service.ts
ewelinagr: good idea, thanks! Will change this

src/app/services/transmart-services/transmart-external-job-resource.service.ts
ewelinagr: true, I meant mocked call comparing to the transmart-resource-service, maybe at some point there will be such a call in tm-packer. Will change this comment.

src/app/services/resource.service.ts
ewelinagr: good point!

README.md
```diff
@@ -135,14 +135,17 @@ Supported properties in the `config.*.json` files:
...
 | `instant-counts-update-1` | `false`   | |
 | `instant-counts-update-2` | `false`   | |
 | `instant-counts-update-3` | `false`   | |
 | `authentication-service-type` | `oidc`  | Authentication service type (`oidc`, `transmart`) |
 | `oidc-server-url`         |           | E.g., `https://keycloak.example.com/auth/realms/{realm}/protocol/openid-connect` |
 | `oidc-client-id`          | `glowingbear-js` | |
+| `use-external-job`        | `false`   | Use an external tool that handles the export. Otherwise the dataView will be used. |
+| `export-data-view`        | `default` | Shape of the export (`default`, `surveyTable`).
+| `export-service-url`      |           | URL of the external tool that handles the export, if the `use-external-job` is set to `true`. |
+| `custom-export-job-name`  |           | Name of the export job that will be triggered in the external tool, if the `use-external-job` is set to `true`. |
```
beaugogh: a bit difficult to understand what this custom-export-job-name is used for, the name will be triggered? doesn't sound like English, refactor the description

src/app/config/app.config.ts
```diff
@@ -97,6 +98,15 @@ export class AppConfig {
         case 'oidc-client-id': {
           value = AppConfig.DEFAULT_OIDC_CLIENT_ID; break;
         }
+        case 'use-external-job': {
+          value = AppConfig.DEFAULT_USE_EXTERNAL_JOB; break;
+        }
+        case 'export-service-url': {
+          throw Error('The Export Service URL is unspecified in the configuration.')
```
beaugogh: this is odd, doesn't seem right to define a config but throw an error here

src/app/config/config.dev.json
```diff
@@ -5,7 +5,10 @@
   "app-version": "0.0.1-SNAPSHOT",
   "doc-url": "https://glowingbear.app",
   "autosave-subject-sets": false,
+  "use-external-job": false,
   "export-data-view": "dataTable",
+  "export-service-url": "http://localhost:8999",
+  "custom-export-job-name": "patient_diagnosis_biosource_biomaterial_export",
```
beaugogh: strange that we need to put custom export job name in config

src/app/models/transmart-models/transmart-external-job.ts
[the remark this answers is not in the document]
beaugogh: Are we deprecating the internal export job? or replacing it with the external one

src/app/app.component.spec.ts
```diff
@@ -46,6 +46,8 @@ import {GrowlModule} from 'primeng/growl';
 import {GbMainModule} from './modules/gb-main-module/gb-main.module';
 import {MessageHelper} from './utilities/message-helper';
 import {of as observableOf} from 'rxjs';
+import {TransmartExternalJobResourceService} from './services/transmart-services/transmart-external-job-resource.service';
```
beaugogh: is it necessary to create a separate external export service here?

beaugogh:
seems the tests are failing

README.md
beaugogh: `{ 
	"export-mode": {
		"mode": "packer",
	    "export-url": "http://foo.bar",
	    "data-view": "patient-dignosis"
	  }
}

=======

{ 
	"export-mode": {
		"mode": "transmart",
	    "data-view": "surveyTable"
	  }
}`

beaugogh: @ewelinagr I edited the config json ⬆

beaugogh:
closing this one in favour of #458

PR #460, opened 2018-10-31
add checkboxes to variables when in export

beaugogh:
TMT-568

### DIFF
--- a/src/app/modules/gb-side-panel-module/accordion-components/gb-variables/gb-variables.component.ts
+++ b/src/app/modules/gb-side-panel-module/accordion-components/gb-variables/gb-variables.component.ts
@@ -51,7 +51,7 @@ export class GbVariablesComponent implements OnInit {
         }
       });
     });
-    this.checkAllText = `${numSelected} variables selected`;
+    this.checkAllText = numSelected > 1 ? `${numSelected} variables selected` : `${numSelected} variable selected`;
   }
 
   checkAll(b: boolean) {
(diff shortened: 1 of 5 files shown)


src/app/modules/gb-side-panel-module/accordion-components/gb-variables/gb-variables.component.ts
```diff
@@ -32,7 +40,23 @@ export class GbVariablesComponent implements OnInit {
         this.categorizedVariables.get(variable.type).push(variable);
       }
     });
-    console.log('categorized variables: ', this.categorizedVariables);
+  }
+
+  onCheckAll(e) {
+    this.checkAllText = e ? 'All selected' : 'None selected';
```
ewelinagr: Having "All selected" or "None selected" test above the variables is a bit confusing, when you select/deselect something not using `CheckAll` checkbox, but a checkbox next to a specific variable. Maybe better to have a counter there as well as a third option: "`x` selected" or at least hide the text at all, when not all/none is selected.

beaugogh: ah, right, never thought of it. will change it!

src/app/modules/gb-side-panel-module/accordion-components/gb-variables/gb-variables.component.ts
```diff
@@ -40,19 +39,28 @@ export class GbVariablesComponent implements OnInit {
...
+  updateCheckAllText() {
+    let numSelected = 0;
+    this.categorizedVariables.forEach((list: Array<Concept>) => {
+      list.forEach((c: Concept) => {
+        if (c.selected) {
+          numSelected++;
+        }
+      });
+    });
+    this.checkAllText = `${numSelected} variables selected`;
```
ewelinagr: maybe variable(s)? to fit also numSelected = 1 :)

beaugogh: good point!

PR #461, opened 2018-10-31
adjust the scope of variables according to cohort selection

beaugogh:


src/app/services/cohort.service.ts
```diff
@@ -34,18 +38,25 @@ export class CohortService {
   private _currentCohort: Cohort;
   // The list of cohorts saved by the user
   private _cohorts: Cohort[] = [];
-  // The total numbers of subjects & observations
+  // The total numbers of subjects & observations for the current cohort
   private _totalCounts: CountItem;
```
ewelinagr: maybe it's better to rename it to sth like `_currentCohotCounts` then? `totalCounts` and `allCounts` are a bit confusing now

beaugogh: true, will do that

beaugogh: I removed the totalCounts all together, since I just realized it is not used.

PR #464, opened 2018-11-02
Fix "Include measurement date columns" button for surveyTable export

ewelinagr:
TMT-498

### DIFF
--- a/src/app/modules/gb-export-module/gb-export.component.ts
+++ b/src/app/modules/gb-export-module/gb-export.component.ts
@@ -25,7 +25,7 @@ export class GbExportComponent implements OnInit, OnDestroy {
 
   constructor(private appConfig: AppConfig,
               private exportService: ExportService,
-              public resourceService: ResourceService) {
+              private resourceService: ResourceService) {
   }
 
   ngOnInit() {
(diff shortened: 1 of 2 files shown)


src/app/modules/gb-export-module/gb-export.component.ts
```diff
@@ -66,15 +66,19 @@ export class GbExportComponent implements OnInit, OnDestroy {
...
   get isExternalExportAvailable(): boolean {
     return this.appConfig.getConfig('export-mode')['name'] !== 'transmart';
   }
 
+  get isTransmartSurveyTableDataView(): boolean {
+    let exportMode =  this.appConfig.getConfig('export-mode');
+    return exportMode['name'] === 'transmart' && exportMode['data-view'] === 'surveyTable';
+  }
+
+  get isTransmartDateColumnIncluded(): boolean {
```
beaugogh: perhaps whether to include dateColumn should also be assessed via config, just like isTransmartSurveyTableDataView

ewelinagr: hmm, but this is now defined via checkbox, which allows users to decide whether to include it or not. I guess what is missing here is a setter.

ewelinagr: it's added now

PR #465, opened 2018-11-05
refactor export flags

beaugogh:


src/app/modules/gb-side-panel-module/accordion-components/gb-tree-nodes/gb-tree-nodes.component.ts
```diff
@@ -225,11 +225,12 @@ export class GbTreeNodesComponent implements OnInit, AfterViewInit, AfterViewChe
             if (subResult.hasMatching) {
               result.hasMatching = true;
               if (this.numExpandedNodes < this.maxNumExpandedNodes) {
-                node['expanded'] = true;
+                expanded = true;
```
beaugogh: somehow ts 3 doesn't allow assigning a boolean value twice, so this change

src/app/services/transmart-services/transmart-resource.service.ts
```diff
@@ -446,11 +449,16 @@ export class TransmartResourceService {
         }));
     } else {
       let transmartTableState: TransmartTableState = null;
```
ewelinagr: this variable can go to the `else {` part below then, but it's really minor

PR #467, opened 2018-11-05
fix export name issue

beaugogh:


ewelinagr:
TMT-544 was also about renaming the `Time` column, right?

src/app/utilities/format-helper.ts
```diff
@@ -41,6 +41,11 @@ export class FormatHelper {
     return 'now';
   }
 
+  public static formatDateString(s: string): string {
+    // TODO: make it more readable, e.g. 2018-10-23T11:26:02Z -> 2018-10-23, 11:26:02
```
ewelinagr: are you still working on this? Or it was not meant to be a part of this fix?

beaugogh: actually, this is sth. extra, I would say it doesn't belong to this fix, just sth. we can improve in future

beaugogh:
@ewelinagr indeed, will change the name of time column

PR #468, opened 2018-11-05
fix query restoring issue - pedigree flags & observation date

beaugogh:
TMT-542

### DIFF
--- a/src/app/utilities/transmart-utilities/transmart-constraint-mapper.ts
+++ b/src/app/utilities/transmart-utilities/transmart-constraint-mapper.ts
@@ -203,7 +203,6 @@ export class TransmartConstraintMapper {
     let constraint: Constraint;
     constraint = new PedigreeConstraint(constraintObject['relationTypeLabel']);
     (<PedigreeConstraint>constraint).biological = constraintObject['biological'];
-    (<PedigreeConstraint>constraint).symmetrical = constraintObject['symmetrical'];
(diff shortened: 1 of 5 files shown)


src/app/utilities/transmart-utilities/transmart-constraint-mapper.ts
```diff
@@ -204,6 +204,7 @@ export class TransmartConstraintMapper {
     constraint = new PedigreeConstraint(constraintObject['relationTypeLabel']);
     (<PedigreeConstraint>constraint).biological = constraintObject['biological'];
     (<PedigreeConstraint>constraint).symmetrical = constraintObject['symmetrical'];
+    (<PedigreeConstraint>constraint).shareHousehold = constraintObject['shareHousehold'];
```
ewelinagr: I am not sure if this is the only place, where some properties of PedigreeConstraint are skipped, e.g.: 

- in this mapper the `description` is not mapped
- [visitPedigreeConstraint(constraint: PedigreeConstraint](https://github.com/thehyve/glowing-bear/blob/dev/src/app/utilities/transmart-utilities/transmart-constraint-serialiser.ts#L240)) -> the `symetrical` is not set there
- [loadPedigrees()](https://github.com/thehyve/glowing-bear/blob/dev/src/app/services/constraint.service.ts#L130) -> `shareHousehold` is not set there

Could you have a look? I am not sure if they should be set, but maybe worth to double check

PR #470, opened 2018-11-08
 refactor export job model

beaugogh:
I ended up refactoring the export job model to distinguish it from the transmart export models. The last update time is formatted in the same way as the date string in subscription


src/app/utilities/transmart-utilities/transmart-packer-mapper.ts
```diff
@@ -20,20 +21,20 @@ const statusDictionary: { [status: string]: string | undefined } = {
 
 export class TransmartPackerMapper {
 
-  static mapCustomExportJob(job: TransmartPackerJob): ExportJob {
-    let ej = new ExportJob();
+  static mapCustomExportJob(job: TransmartPackerJob): TransmartExportJob {
```
ewelinagr: hmm, so it's mapped from `TransmartPackerJob` to `TransmartExportJob` and then to `ExportJob`? Is this necessary?

beaugogh: for the architecture that we have now, yes, since transmartResourceSerivce now acts as a switch, and the export job it outputs would be 'transmartExportJob'. but indeed, in future if we are adopting essentially three endpoints, 1.transmart, 2.packer and 3.fractalis, we need to rethink the way we organize resourceService, fractalisService and transamrt-related services.

PR #478, opened 2018-11-20
Add export job name validation.

gijskant:


src/app/modules/gb-export-module/gb-export.component.html
```diff
@@ -7,46 +7,80 @@ <h5>Create new export for the currently selected data</h5>
...
+                 name="exportJobNameInput"
+                 #exportJobNameInput="ngModel"
+                 placeholder="Export job name"
+                 (drop)="onExportJobNameInputDrop($event)"
+                 required
+                 minlength="1"
+                 maxlength="200"
+                 gbNonEmpty
+                 [gbExcludeValues]="jobNames"
+                 [(ngModel)]="exportJobName">
```
ewelinagr: Maybe it would be useful to validate also the special characters? Currently when I use characters like `+` or `#` etc. they are just replaced by white spaces. When I use only them - the whole name is replaced with job id and event though the job is created I the error see message: "A job name is required. "
This is unlikely that the user will provide such a name, but possible.

src/app/services/export.service.ts
```diff
@@ -246,23 +246,23 @@ export class ExportService {
     }
     // 2. Validate if job name is not duplicated
     for (let job of this.exportJobs) {
-      if (job['jobName'] === name) {
+      if (job.name === name) {
         const summary = 'Duplicate job name, choose a new name.';
```
ewelinagr: is this validation still needed? Looks like the name is validated twice now

gijskant: Yes, most of it can go now.

src/app/modules/gb-export-module/gb-export.component.html
gijskant: Interesting, do you know where that happens?

ewelinagr: Looks like `#` (and every character after #) is removed automatically from the request query string, but others not sure

ewelinagr: but maybe it's better to merge this PR and report special characters validation as a separate issue?

gijskant: It should be fixed now by using [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) on the job name.

PR #481, opened 2018-12-06
Improve performance when using exclusion criteria on large data sets.

gijskant:


src/app/services/transmart-services/transmart-resource.service.ts
```diff
@@ -255,24 +271,24 @@ export class TransmartResourceService {
...
     return new Promise<any>((resolve, reject) => {
       if (this.autosaveSubjectSets) {
-        if (totalCountItem) {
-          this.inclusionCounts.patientCount = totalCountItem.patientCount + this.exclusionCounts.patientCount;
-          this.inclusionCounts.observationCount = totalCountItem.observationCount + this.exclusionCounts.observationCount;
-          resolve(true);
+        if (!totalCountItem) {
+          return reject('Missing transmart total-count item to calculate inclusion counts.');
+        }
+        this.inclusionCounts.patientCount = totalCountItem.patientCount - this.exclusionCounts.patientCount;
```
ewelinagr: hmm, are you sure about the subtraction character? Previously `total = inclusion - exclusion`, so `inclusion = total + exclusion` (there was a `+` here).

gijskant: You are correct, I will fix it.

gijskant:
Still some weird issues with 'step 2' when using `autosave-subject-sets`.

ewelinagr:
Looks good! Though the rabasing of fractalis-integration branch with dev might be more challenging now :)

PR #482, opened 2018-12-06
Revert and refactor import criteria functionality

ewelinagr:
[TMT-555](https://jira.thehyve.nl/browse/TMT-555) & [TMT-556](https://jira.thehyve.nl/browse/TMT-556)

TMT-556 is not fully completed - it depends on the variables panel that will be changed in [TMT-631](https://jira.thehyve.nl/browse/TMT-631), so the variables import will be changed and finished accordingly.

### DIFF
--- a/src/app/utilities/file-import-helper.ts
+++ b/src/app/utilities/file-import-helper.ts
@@ -1,0 +1,37 @@
+import {MessageHelper} from './message-helper';
+
+export class FileImportHelper {
+
+  static importCriteria(elementId: string, reader: FileReader, isUploadListenerNotAdded: boolean) {
+    let uploadElm = document.getElementById(elementId);
+    if (isUploadListenerNotAdded) {
+      uploadElm.addEventListener('change', (event) => this.fileUpload(event, reader), false);
+    }
+    // reset the input path so that it will take the same file again
+    uploadElm['value'] = '';
+    uploadElm.click();
+  }
+
+  static fileUpload(event, reader: FileReader) {
+    MessageHelper.alert('info', 'File is being processed, waiting for response.');
+    let file = event.target.files[0];
+    reader.readAsText(file);
+  }
+
+  static getFile(elementId: string) {
+    return (<HTMLInputElement>document.getElementById(elementId)).files[0];
+  }
+
+  static isTextFile(file: File): boolean {
+    return file.type === 'text/plain' ||
+      file.type === 'text/tab-separated-values' ||
+      file.type === 'text/csv' ||
+      (file.type === '' && file.name.split('.').pop() !== 'json');
+  }
+
+  static isJsonFile(file: File): boolean {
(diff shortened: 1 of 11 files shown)


src/app/utilities/file-import-helper.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+    // reset the input path so that it will take the same file again
+    uploadElm['value'] = '';
+    uploadElm.click();
+  }
+
+  static fileUpload(event, reader: FileReader) {
+    MessageHelper.alert('info', 'File is being processed, waiting for response.');
+    let file = event.target.files[0];
+    reader.readAsText(file);
```
beaugogh: also a message telling it's finished?

src/app/services/tree-node.service.ts
```diff
@@ -357,15 +357,15 @@ export class TreeNodeService {
    * @param {string[]} items
    * @param {string[]} paths
    */
-  public convertItemsToPaths(nodes: TreeNode[], items: string[], paths: string[]) {
+  public static convertItemsToPaths(nodes: TreeNode[], items: string[], paths: string[]) {
```
beaugogh: why is this changed to static?

beaugogh:
A general remark: we had issues with these two imports consistently, it would be better to add some integration or e2e tests that also cover edge cases

src/app/utilities/file-import-helper.ts
ewelinagr: I can add one, but there are already 3 messages showing up, when successfully loading file - 1 on start and 2 when executing restoreCohort. Maybe it would be too much?

ewelinagr:
@beaugogh you're right about the tests, I created an extra ticket: https://jira.thehyve.nl/browse/TMT-636

src/app/utilities/file-import-helper.ts
beaugogh: I would put up a message at the beginning of the import (inside the importCriteria method of gb-cohort-selection.component.ts) saying it has started, and another message when onLoad is finished, i.e. in handleVariablesFileUploadEvent saying it is finished, and in case error, say there is error (which we already have)

src/app/utilities/file-import-helper.ts
```diff
@@ -0,0 +1,37 @@
+import {MessageHelper} from './message-helper';
+
+export class FileImportHelper {
+
+  static importCriteria(elementId: string, reader: FileReader, isUploadListenerNotAdded: boolean) {
+    let uploadElm = document.getElementById(elementId);
+    if (isUploadListenerNotAdded) {
+      uploadElm.addEventListener('change', (event) => this.fileUpload(event, reader), false);
```
beaugogh: `this.fileUpload` -> FileImportHelper.fileUpload

PR #487, opened 2018-12-12
Add double-view mode to the variables panel

ewelinagr:


ewelinagr:
@gijskant @beaugogh I am still extending tests, but you could already have a look at the code.

src/app/models/chart-models/chart.ts
```diff
@@ -53,11 +53,13 @@ export class Chart {
   }
 
   get categoricalVariables(): Concept[] {
-    return this.variables.filter( variable => variable.type !== ConceptType.NUMERICAL);
+    return this.variables.filter(variable =>
+      [ConceptType.CATEGORICAL, ConceptType.DATE].indexOf(variable.type) > -1);
```
gijskant: In TypeScript you can use `list.includes(value)` instead of `list.indexOf(value) !== -1`.

src/app/services/constraint.service.ts
```diff
@@ -124,6 +131,8 @@ export class ConstraintService {
   // [pDroppable] in gb-cross-table.component
   public variablesDragDropScope = 'PrimeNGVariablesDragDropContext';
 
+  private _variablesViewMode: VariablesViewMode;
```
gijskant: At first sight, this does not look like the right place to store this view mode. But then, it looks like this service already has gotten out of hand, it has too many responsibilities already.

src/app/services/constraint.service.ts
```diff
@@ -407,10 +417,33 @@ export class ConstraintService {
...
+      treeNodeType === 'DATE' ||
+      treeNodeType === 'HIGH_DIMENSIONAL' ||
+      treeNodeType === 'TEXT')
+  }
+
   /**
    * Generate the constraint based on the variables selected in the Variables panel
    */
   public variableConstraint(): Constraint {
+    if (this.variablesViewMode === VariablesViewMode.TREE_VIEW) {
```
gijskant: Is it not possible to store the selected variables in one place? Then also your selection is maintained when switching view mode.

src/app/services/fractalis.service.ts
```diff
@@ -66,13 +66,18 @@ export class FractalisService {
...
         };
+        const type = this.mapConceptTypeToFractalisVariableType(variable.type);
+        if (!type) {
+          console.error(`Failed to load variable: ${variable.code}. Invalid type: ${variable.type}.`);
+          return;
+        }
         const descriptor = {
-          constraint: JSON.stringify(constraint),
-          data_type: variable.type === ConceptType.NUMERICAL ? 'numerical' : 'categorical',
+          constraint: JSON.stringify(fractalisConstraint),
```
gijskant: `JSON.stringify` is not needed.

PR #489, opened 2018-12-17
Add variables upload status validation

ewelinagr:
[TMT-627](https://jira.thehyve.nl/browse/TMT-627)

### DIFF
--- a/src/app/services/fractalis.service.ts
+++ b/src/app/services/fractalis.service.ts
@@ -9,6 +9,8 @@ import {Concept} from '../models/constraint-models/concept';
 import {ConceptType} from '../models/constraint-models/concept-type';
 import {ConstraintService} from './constraint.service';
 import {AppConfig} from '../config/app.config';
+import {FractalisData} from '../models/fractalis-models/fractalis-data';
+import {FractalisEtlState} from '../models/fractalis-models/fractalis-etl-state';
 
 @Injectable({
   providedIn: 'root'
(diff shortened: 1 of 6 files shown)


src/app/services/fractalis.service.ts
```diff
@@ -174,6 +179,45 @@ export class FractalisService {
     this.variablesInvalid = false;
   }
 
+  public validateVariableUploadStatus(variable: Concept): Promise<boolean> {
+    return new Promise<boolean>((resolve, reject) => {
+      this.getLoadedVariables().then(dataObj => {
+        let message = '';
+        if (dataObj) {
+          let fractalisDataList: FractalisData[] = FractalisService.dataObjectToFractalisDataList(dataObj);
+          let fractalisVars = fractalisDataList.filter(data => data.label === variable.code);
```
ewelinagr: @gijskant this one links the variables that were loaded into fractalis with GB variables based on the concept code (which is a variable label in fractalis). There is no guarantee that the label is unique (though we can assume that the concept code is unique, right?). There is an option to base this on fractalis unique task_id, but then mapping between concepts and task_ids has to be stored here and also there would be a small change in fractalis required to return the task_id in the POST /data call. Do you think this might would be required?

gijskant: Indeed, concept code is better (and unique), probably the only good identifier to use. Using the name makes it hard to check if the variable has already been loaded (name colisions). However, it is not very descriptive. Can we add a legend that adds a descriptive label for the variables?

ewelinagr: You mean for variables in fractalis or the variables panel in GB? If you meant the panel, both views still display the names of variables, the grouped (categorized) view probably should display also the concept code

PR #490, opened 2018-12-17
Fractalis ui improvements

ewelinagr:
[TMT-628](https://jira.thehyve.nl/browse/TMT-628)
[TMT-632](https://jira.thehyve.nl/browse/TMT-632)

- Remove angular-gridster2 
- Add slider to adjust chart size
- Add styling improvements for cohort, analysis and export tabs

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
     "@angular/platform-server": "^7.0.3",
     "@angular/router": "^7.0.3",
     "@types/file-saver": "^1.3.0",
-    "angular-gridster2": "^7.0.0",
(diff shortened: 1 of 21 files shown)


gijskant:
Can you also remove the gridster dependency from package.json (and add the yarn.lock after running yarn?)

gijskant:
Hmm, I did not expect all integrity checksums to be removed from yarn.lock, which yarn version are you using? Could be that something in yarn has changed that this is not needed?

ewelinagr:
@gijskant I have yarn 1.9.4, not sure why this happened... I can try with yarn 1.10

PR #493, opened 2018-12-17
Change chart validation to support the survival plot.

gijskant:
- [x] Need to add tests to gb-fractalis-chart.component.spec.ts.


src/app/modules/gb-analysis-module/fractalis-components/gb-fractalis-control/fractalis-chart-variable-mapper.ts
```diff
@@ -0,0 +1,150 @@
...
+   * @return {ValidationError[]}
+   */
+  static validateMapping(mapping: Map<string, string[]>, specification: FractalisChartDescription): ValidationError[] {
+    const errors: ValidationError[] = [];
+    for (let key of this.getParameterKeys(specification)) {
+      const label = specification[key].label;
+      let variableTypeSpecification: FractalisChartVariable = specification[key];
+      let declaredVariables = mapping.get(key);
+      const minLength = variableTypeSpecification.minLength;
+      const maxLength = variableTypeSpecification.maxLength;
```
ewelinagr: It seems like something is missing in this validation logic. If `minLength` and `maxLength` are not specified, for me it means that variable type is not supported for the current chart. Currently this is not validated, e.g. for volcano plot or heatmap I can add categorical variables, even though the descriptions of these charts specify only the numVars.

src/app/services/fractalis.service.ts
```diff
@@ -134,8 +137,25 @@ export class FractalisService {
     }
   }
 
-  setChart(chartId: string): object {
-    return this.F.setChart(this.selectedChartType, chartId);
+  initChart(chartType: ChartType, chartId: string): Observable<FractalisChart> {
+    const chartObject = this.F.setChart(chartType, chartId);
+    const chartSubject = new BehaviorSubject<FractalisChart>({
```
ewelinagr: nice!

ewelinagr:
I really like the refactoring of the chart validation! It makes the gb-fractalis-chart component much cleaner, also covers Bo's refactoring suggestion described in [TMT-627 ](https://jira.thehyve.nl/browse/TMT-627)

src/app/modules/gb-analysis-module/fractalis-components/gb-fractalis-control/fractalis-chart-variable-mapper.ts
ewelinagr: works fine now

PR #494, opened 2018-12-17
Update fractalis auth parameters

ewelinagr:


src/app/services/fractalis.service.ts
```diff
@@ -52,12 +52,19 @@ export class FractalisService {
...
       handler: 'transmart',
       dataSource: this.appConfig.getConfig('fractalis-datasource-url'),
       fractalisNode: this.appConfig.getConfig('fractalis-url'),
       getAuth() {
-        return {token: token}
+        return {
+          token: token,
+          authServiceType: 'oidc',
+          oidcClientId: oidcClientId,
+          oidcServerUrl: oidcServerUrl,
```
gijskant: Having the authentication server specified by the client makes it required to be part of the session key, i.e., when the client specifies a different url (could be one setup by an attacker, to verify a token created by the attacker), a new session is created.

gijskant: Ah, so this auth object is only sent when doing a new data request to the backend (`POST /data`), not when fetching data. I guess we should also sent an `Authorization` header for other requests if we want the Fractalis backend to be able to verify other requests.

ewelinagr: yes, but this will require some changes to fractal.js front-end code too. We could add it for our deployment (generate our own fractal.js) or wait for the next version of fractalis front-end, what do you think?

gijskant: Let's merge this now and put those improvements on the list later this week.

PR #495, opened 2018-12-19
Upgreade cypress to 3d version

forus:


ewelinagr:
why did you delete the yarn.lock? 
the build is failing probably because of that

forus:
@ewelinagr `.lock` file did look like system file to me. I've removed that commit and rebased change on recent dev. Please have another look.

PR #497, opened 2018-12-21
Add possibility to disable fractalis analysis

ewelinagr:
TMT-610

TODO
- [ ] Add unit tests


gijskant:
It seems that disabling works, but when enabled, I get no chart types (except cross table). It does perform many `GET /data` calls in the background.

Related issues, but not caused by this PR:
- Cross table variable selection only works from categorised view, not from the tree view
- Should the variables panel be visible when in the cohort selection tab?

ewelinagr:
@gijskant thanks for noticing! It was my silly mistake, it is fixed now. 

Regarding the related issues - first one will be covered by [TMT-663](https://jira.thehyve.nl/browse/TMT-663). Regarding the second one, some time ago we decided to keep it to have a preview of variables belonging to a selected cohort, but indeed it is possible in other tabs too, so we could discuss this again.

beaugogh:
Just some general comments about the ticket itself, which seems to need more details, see https://jira.thehyve.nl/browse/TMT-610

PR #499, opened 2018-12-31
Add material-design icons to package.json

ewelinagr:


ewelinagr:
@beaugogh That is true, but for some reason the icons are not visible on https://glowingbear-dev.thehyve.net. Adding it like this was recommended as a solution.

PR #511, opened 2019-01-10
refactor fractalis variable loading

beaugogh:


src/app/models/fractalis-models/fractalis-variables-status.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Copyright 2017 - 2018  The Hyve B.V.
```
ewelinagr: 2019 :)

src/app/services/fractalis.service.ts
```diff
@@ -79,50 +85,87 @@ export class FractalisService {
...
-    return this.F.clearCache();
+  public clearCache() {
+    this.isClearingCache = true;
+    this.F.clearCache()
+      .then(res => {
+        this.isClearingCache = false;
+      })
+      .catch(error => {
+        console.error(`Error clearing Fractalis cache: ${error}`);
+        this.clearCache();
```
ewelinagr: Maybe better to have some max retrial number not to stuck in the loop

PR #512, opened 2019-01-14
add setSubsets in fractalis service

beaugogh:


src/app/services/fractalis.service.ts
```diff
@@ -157,9 +175,30 @@ export class FractalisService {
...
+  private configSubsetsUpdate() {
+    this.cohortService.cohortsUpdated.asObservable()
+      .subscribe((cohorts: Cohort[]) => {
+        let subjectCalls = [];
+        cohorts
+          .filter((cohort: Cohort) => {
+            return cohort.selected;
+          })
+          .forEach((cohort: Cohort) => {
+            subjectCalls.push(this.resourceService.getSubjects(cohort.constraint));
```
gijskant: Does this fetch patient ids for all cohorts whenever one of the cohorts has changed or selected?

src/app/services/fractalis.service.ts
```diff
@@ -157,9 +175,30 @@ export class FractalisService {
...
+          });
+        let idSets = [];
+        forkJoin(subjectCalls)
+          .subscribe(res => {
+            for (let i = 0; i < subjectCalls.length; i++) {
+              const ids = res[i].map((subject: TransmartPatient) => {
+                return subject.id;
+              });
+              idSets.push(ids);
```
ewelinagr: Looks like this way there may be some duplications of ids on the `idSets` list

src/app/services/fractalis.service.ts
beaugogh: yes, whenever one of the cohorts is changed, the cohortsUpdated subject will emit event to inform its subscribers, in this case, the only subscriber here in fractalis service, catches it and does subsets update

src/app/modules/gb-side-panel-module/accordion-components/gb-cohorts/gb-cohorts.component.ts
```diff
@@ -217,7 +217,8 @@ export class GbCohortsComponent implements OnInit {
 
   onCohortCheckClick(e: MouseEvent) {
     e.stopPropagation();
-    this.cohortService.updateAll();
+    this.cohortService.cohortsUpdated.next(this.cohortService.cohorts);
+    this.cohortService.updateCountsWithAllCohorts();
```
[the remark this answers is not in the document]
beaugogh: the cohortsUpdated subject will do the update inside updateCountsWithAllCohorts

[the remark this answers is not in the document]
beaugogh: somehow the change is not reflected here, see https://github.com/thehyve/glowing-bear/blob/mcohorts/src/app/modules/gb-side-panel-module/accordion-components/gb-cohorts/gb-cohorts.component.ts

PR #515, opened 2019-01-22
add cypress tests for boxplot + refactoring

beaugogh:


cypress/integration/fractalis-charts.feature
```diff
@@ -0,0 +1,18 @@
+Feature: Create various Fractalis charts
+
+  Scenario: Create a box-plot with a numerical variable from the Category View with one cohort
+    Given I am on the Analysis tab
+    When I select a box plot
+    And I drag numerical_11 variable from the Category View to variable drop zone
+    And I add the chart
```
gijskant: Looks like the step definition for this step is missing.

beaugogh: good catch :)

PR #518, opened 2019-02-05
Use gb-backend for user-queries related calls 

ewelinagr:
[TMT-676](https://jira.thehyve.nl/browse/TMT-676)


ewelinagr:
Refactoring of http services will be covered by [TMT-737](https://jira.thehyve.nl/browse/TMT-737)

src/app/services/http/gb-backend-http.service.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * save a new query
+   * @param {GbBackendQuery} query
+   * @returns {Observable<GbBackendQuery>}
+   */
+  saveQuery(query: GbBackendQuery): Observable<GbBackendQuery> {
+    const urlPart = `queries`;
+    const queryBody = {};
```
gijskant: Why not just send the query object directly?

gijskant: I don't understand why if would be different, actually.

ewelinagr:
Since this is a major change, before merging this PR better to merge other PRs first and tag a version/create a release.

README.md
```diff
@@ -143,6 +143,7 @@ Supported properties in the `config.*.json` files:
 | `api-url`                 |           | URL of the TranSMART API to connect to. |
 | `api-version`             | `v2`      | TranSMART API version. Only `v2` is supported. |
 | `app-url`                 |           | URL where the Glowing Bear is accessible for the user.|
+| `gb-backend-url`          |           | URL of the Gb-backend application to connect to for cohorts handling. |
```
beaugogh: is this to replace api-url? if so, remove api-url completely? or point api-url to gb-backend-url?

ewelinagr: actually gb-backend is additional application with user queries only. The whole query related functionality (calls starting with '/queries') was moved from transmart to gb-backend, but the other calls will be still handled in transmart. Maybe we could rename the api-url to transmart-url, or something like this, since it's no longer the only API gb is using.

beaugogh: this seems to be an effort of better organising/ modularising transmart code, however, from glowing bear perspective, there is just one backend, which is capable of handling all the api calls, separating /queries calls from other calls seems arbitrary, front-end wise. It is better to have a unified frontend also for the reason that glowing bear is intended to be a cohort selection tool not just for transmart, but also for other backends, having two backends for transmart calls is not that sensible and is obstructive for future extension.

beaugogh: how much effort does it require for gb-backend to relay non queries calls to transmart? so that gb-backend is the only cohort-selection/export backend for glowingbear

beaugogh: never mind, this seems to be too much to change, I am fine with the current choice

src/app/services/http/gb-backend-http.service.ts
ewelinagr: you're right, this mapping seems not to be needed, I thought createDate/updateDate may cause some issues, but it works well. It's updated now.

PR #519, opened 2019-02-06
Fix survey table export.

gijskant:
Fixes [NTRUAT-212](https://jira.thehyve.nl/browse/NTRUAT-212).

Needs to be applied on dev as well, if this solution works.

### DIFF
--- a/src/app/services/transmart-services/transmart-resource.service.ts
+++ b/src/app/services/transmart-services/transmart-resource.service.ts
@@ -269,7 +269,6 @@ export class TransmartResourceService {
       if (this.autosaveSubjectSets) {
         this.savePatientSet('temp', constraint)
           .subscribe((subjectSet: SubjectSet) => {
-            this.subjectSetConstraint.id = subjectSet.id;
             this.exclusionCounts.patientCount = subjectSet.setSize;
             this.exclusionCounts.observationCount = -1;
             resolve(true);


src/app/services/transmart-services/transmart-resource.service.ts
```diff
@@ -269,7 +269,6 @@ export class TransmartResourceService {
       if (this.autosaveSubjectSets) {
         this.savePatientSet('temp', constraint)
           .subscribe((subjectSet: SubjectSet) => {
-            this.subjectSetConstraint.id = subjectSet.id;
```
beaugogh: why is the id not assigned to the subject constraint?

gijskant: The constraint that is being persisted at this point is the exclusion constraint. This subjectSetConstraint object is used, e.g., to specify the cohort to be exported, so only the subject set for the combination of inclusion and exclusion criteria should be saved there.

gijskant: Currently, when updating the cohort, two processes will update the same field (`subjectSetConstraint.id`): here for the exclusion constraint, somewhere else for the combination.

beaugogh: indeed, maybe should have created three patient sets, one combined, one inclusive, one exclusive, but since we are to remove exclusion constraint, all this trouble would go away :D

beaugogh: also currently it seems that we only save patient sets, not using the ones already saved, something to improve

PR #520, opened 2019-02-12
sync variable selection in tree view and category view

beaugogh:


src/app/services/constraint.service.ts
```diff
@@ -119,8 +120,11 @@ export class ConstraintService {
   * when user defines a cohort or a combination of cohorts.
   */
   private _variables: Concept[] = [];
+  private _categorizedVariables: Array<CategorizedVariable> = [];
```
gijskant: Please separate cohort selection and variable selection into separate services.

src/app/services/constraint.service.ts
```diff
@@ -579,40 +641,31 @@ export class ConstraintService {
...
   importVariablesByNames(names: string[]) {
-    if (this.variablesViewMode === VariablesViewMode.TREE_VIEW) {
-      return this.treeNodeService.selectVariablesTreeNodesByNames(this.treeNodeService.variablesTreeData, names);
-    } else if (this.variablesViewMode === VariablesViewMode.CATEGORIZED_VIEW) {
-      return this.selectVariablesByProperty(names, 'name');
-    }
+    // update the selected tree nodes in gb-variables
+    this.treeNodeService.selectVariablesTreeDataByFields(
+      this.treeNodeService.variablesTreeData, names, ['metadata', 'item_name']);
+    // dispatch the event telling its subscribers that the selected tree nodes have been updated
```
gijskant: That should be the responsibility of `treeNodeService.selectVariablesTreeDataByFields`, right?

src/app/services/mocks/constraint.service.mock.ts
```diff
@@ -63,4 +66,32 @@ export class ConstraintServiceMock {
   public identifyDraggedElement(): Concept {
     return new Concept();
   }
+
+  checkAllVariables(b: boolean) {
```
gijskant: `checkAllVariables` is the wrong name, should be something like `setVariableSelection`.

src/app/services/mocks/tree-node.service.mock.ts
```diff
@@ -16,17 +16,13 @@ export class TreeNodeServiceMock {
...
-  private _selectedProjectionTreeData: TreeNode[] = [];
-  // the final tree nodes resulted from data selection
-  private _finalTreeNodes: TreeNode[] = [];
+  private _variablesTreeData: TreeNode[] = [];
 
   public selectedTreeNode;
   public treeNodeCallsSent = 0; // the number of tree-node calls sent
   public treeNodeCallsReceived = 0; // the number of tree-node calls received
   public treeNodesUpdated: Subject<boolean> = new Subject<boolean>();
+  public selectedVariablesTreeDataUpdated: Subject<TreeNode[]> = new Subject<TreeNode[]>();
```
gijskant: Instead of having variable selection update events in both services (`selectedVariablesTreeDataUpdated` in `TreeNodeService` and `selectedVariablesUpdated` in `ConstraintService`), it is better to have one place (possibly in a variable selection service), have all components update the selection there, and having the tree node service and constraint service (or perhaps categorized variable component) listen to these updates to apply the selection.

src/app/services/constraint.service.ts
beaugogh: cohort selection?

src/app/services/constraint.service.ts
beaugogh: indeed, will change

src/app/services/mocks/constraint.service.mock.ts
beaugogh: will change

PR #521, opened 2019-02-18
Add a panel level negation to the constraint builder

ewelinagr:
[TMT-722](https://jira.thehyve.nl/browse/TMT-722)

### DIFF
--- a/src/app/modules/gb-cohort-selection-module/constraint-components/gb-constraint/gb-constraint.component.ts
+++ b/src/app/modules/gb-cohort-selection-module/constraint-components/gb-constraint/gb-constraint.component.ts
@@ -87,29 +87,12 @@ export class GbConstraintComponent implements OnInit {
     }
   }
 
-  get containerClass(): string {
-    if (this.element.nativeElement.children[0].classList.length === 0) {
-      const containerClassName = (this.constraint.className === 'CombinationConstraint'
-        && (<CombinationConstraint>this.constraint).isRoot) ?
-        'gb-constraint-container-root ' : 'gb-constraint-container';
-
-      let borderClassName = '';
-      if (containerClassName === 'gb-constraint-container') {
-        const depth = this.constraintService.depthOfConstraint(this.constraint);
-        if (depth === 1) {
-          borderClassName = 'gb-constraint-container-border-left-1';
-        } else if (depth === 2) {
-          borderClassName = 'gb-constraint-container-border-left-2';
-        } else if (depth === 3) {
-          borderClassName = 'gb-constraint-container-border-left-3';
-        } else {
-          borderClassName = 'gb-constraint-container-border-left-4';
-        }
-      }
-      return containerClassName + ' ' + borderClassName;
-    } else {
-      return this.element.nativeElement.children[0].classList.value;
-    }
+  get constraintDepth(): number {
+    return ConstraintService.depthOfConstraint(this.constraint);
   }
 
+  get isRootClassLevel(): boolean {
+    return this.constraint.className === 'CombinationConstraint'
+      && (<CombinationConstraint>this.constraint).isRoot;
+  }
 }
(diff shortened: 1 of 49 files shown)


ewelinagr:
Should be merged together with an upcoming PR for TMT-721 - Get rid of exclusion criteria

src/app/utilities/constraint-utilities/abstract-constraint-visitor.ts
```diff
@@ -21,7 +21,17 @@ import {StudyConstraint} from '../../models/constraint-models/study-constraint';
...
+    if (constraint.negated) {
+      let constraintCopy = Object.create(constraint);
+      constraintCopy.negated = false;
+      return new NegationConstraint(constraintCopy);
+    }
+    return constraint;
+  }
+
   visit(constraint: Constraint): T {
+    constraint = AbstractConstraintVisitor.replaceNegationFlagWithNegationConstraint(constraint);
```
gijskant: In true visitor style, you could write
```javascript
if (constraint.negated) {
  let constraintCopy = Object.create(constraint);
  constraintCopy.negated = false;
  return this.visit(NegationConstraint(constraintCopy));
}
```

src/app/modules/gb-cohort-selection-module/constraint-components/gb-constraint/gb-constraint.component.ts
```diff
@@ -48,6 +50,11 @@ export class GbConstraintComponent implements OnInit {
     this.constraintRemoved.emit();
   }
 
+  negate() {
```
gijskant: this method name sounds a bit confusing to me. It updates the constraint negation based on the component slider, right? Is it actually required to save this `excluded` boolean separately, or could the slider also update `constraint.negated` directly?

src/app/services/constraint.service.ts
```diff
@@ -115,11 +115,6 @@ export class ConstraintService {
...
               private studyService: StudyService,
               private resourceService: ResourceService) {
-    // Initialize the root inclusion and exclusion constraints in the 1st step
-    this.rootInclusionConstraint = new CombinationConstraint();
-    this.rootInclusionConstraint.isRoot = true;
-    this.rootExclusionConstraint = new CombinationConstraint();
-    this.rootExclusionConstraint.isRoot = true;
 
     // Initialize the root inclusion and exclusion constraints in the 1st step
     this.rootInclusionConstraint = new CombinationConstraint();
```
beaugogh: so maybe not called rootInclusionConstraint anymore? just rootConstraint

src/app/modules/gb-cohort-selection-module/gb-cohort-selection.module.ts
```diff
@@ -46,7 +46,8 @@ import {GbPedigreeConstraintComponent} from './constraint-components/gb-pedigree
     DropdownModule,
     MessagesModule,
     MultiSelectModule,
-    OverlayPanelModule
+    OverlayPanelModule,
+    InputSwitchModule
```
beaugogh: showing a label next to an input switch would be too much space? especially when the query is big and nested. How about just use a basic toggle button? as in here [https://www.primefaces.org/primeng/#/togglebutton](url)

src/app/services/constraint.service.ts
ewelinagr: this is already renamed in https://github.com/thehyve/glowing-bear/pull/523 - removal of exclusion constraint, I will merge these 2 PRs together soon (I spitted them so it would be easier to review)

src/app/modules/gb-cohort-selection-module/gb-cohort-selection.module.ts
ewelinagr: There will be some styling change added for [TMT-727](https://jira.thehyve.nl/browse/TMT-727), when we introduce the context aware user friendly box messages. With the description in the top part of the box it will be possible to move the input switch up (so it will be aligned with the text). I guess then it will not be a problem that the switch with the label is taking more space. 
Probably better to discuss the styling changes with Elisa.

beaugogh: functionally toggle button and switch input are the same, just the latter occupies less space, always good to save some (precious) space

ewelinagr: it is addressed here: [TMT-766](https://jira.thehyve.nl/browse/TMT-766)

PR #523, opened 2019-02-19
Remove exclusion criteria, refactor counts

ewelinagr:
[TMT-721](https://jira.thehyve.nl/browse/TMT-721)
[TMT-680](https://jira.thehyve.nl/browse/TMT-680)

### DIFF
--- a/cypress/integration/constrain-creation.feature
+++ b/cypress/integration/constrain-creation.feature
@@ -6,7 +6,12 @@ Feature: create constraints by selecting nodes from the tree
     When I use public study 'CATEGORICAL_VALUES ' as a constraint
     Then there are '3' subjects
 
-  Scenario: create a constraint with negation
+  Scenario: create a constraint with concept negation
+    Given I am on the cohort-selection tab
+    When I select study Oracle_1000_Patient but exclude from categorical_10, Stomach, Lung, Head, Liver
+    Then there are '654' subjects
+
+  Scenario: create a constraint with study negation
     Given I am on the cohort-selection tab
     When  I use public study 'EHR' and negation of study 'CATEGORICAL_VALUES' as a constraint
     Then there are '3' subjects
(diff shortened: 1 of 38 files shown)


cypress/integration/constrain-creation.feature
```diff
@@ -6,7 +6,12 @@ Feature: create constraints by selecting nodes from the tree
     When I use public study 'CATEGORICAL_VALUES ' as a constraint
     Then there are '3' subjects
```
beaugogh: looks like the time to correct the file name: constrain-creation -> constraint-creation :)

ewelinagr: ah, didn't even notice that ;) I will rename it!

src/app/services/cohort.service.ts
```diff
@@ -44,60 +46,46 @@ export class CohortService {
...
   private _isUpdatingAll = false;
   // flag indicating if the current cohort constraint in gb-cohort-selection has been changed
   private _isDirty = true;
   private _cohortsUpdated: Subject<Cohort[]> = new Subject<Cohort[]>();
-
+  // indicate when constraints are changed, whether to update counts immediately,
+  // used in gb-constraint-component
+  private _instantCohortCountsUpdate: boolean;
+  // indicate when constraints are changed, whether to update counts immediately,
+  // used in gb-constraint-component
```
beaugogh: same comment as the previous one?

src/app/services/cohort.service.ts
```diff
@@ -189,34 +164,48 @@ export class CohortService {
...
+      let combination: CombinationConstraint = new CombinationConstraint();
+      combination.combinationState = CombinationState.And;
+      combination.mark = ConstraintMark.SUBJECT;
+      this.cohorts.forEach((cohort: Cohort) => {
+        if (cohort.selected) {
+          combination.addChild(cohort.constraint);
+        }
+      });
+      if (this.saveSubjectSetBeforeUpdatingCounts) {
+        this.resourceService.saveSubjectSet('temp', combination).subscribe((subjectSet: SubjectSet) => {
```
beaugogh: if the 'combination' constraint has been used before, there would be no need to save another subjectSet, right? since a subjectSet has been saved previously for that constraint

beaugogh: probably involves a more extensive discussion

PR #525, opened 2019-02-20
fix variable selection bug and add comments to integration test

beaugogh:


src/app/services/variable.service.ts
```diff
@@ -250,6 +255,7 @@ export class VariableService {
         const val = fields.length < 2 ? node[fields[0]] : (node[fields[0]] || {})[fields[1]];
         if (values.includes(val) && !this.selectedVariablesTree.includes(node)) {
           this.selectedVariablesTree.push(node);
+          console.log('apush node', node)
```
gijskant: I guess this one can go?

PR #530, opened 2019-02-25
fix check all checkbox bug in variable selection

beaugogh:


src/app/modules/gb-side-panel-module/accordion-components/gb-variables/gb-variables.component.ts
```diff
@@ -13,26 +21,29 @@ import {VariableService} from '../../../../services/variable.service';
...
               private navbarService: NavbarService) {
-    this.allChecked = true;
     this.isUploadListenerNotAdded = true;
-    this.viewMode = VariablesViewMode.TREE_VIEW;
-    this.availableViewModes = this.listAvailableViewModes();
   }
 
   ngOnInit() {
-    this.checkAllVariables(true);
+    this.availableViewModes = Object.keys(VariablesViewMode).map(c => {
```
forus: Having logic to collect all possible select items in the separate function `listAvailableViewModes()` was not that bad idea.

src/app/modules/gb-side-panel-module/accordion-components/gb-variables/gb-variables.component.ts
```diff
@@ -13,26 +21,29 @@ import {VariableService} from '../../../../services/variable.service';
 })
 export class GbVariablesComponent implements OnInit {
 
-  public allChecked: boolean;
+  private _allChecked: boolean;
```
forus: do you need this field? it seems to me like setter and getter only will do.

src/app/modules/gb-side-panel-module/accordion-components/gb-variables/gb-variables.component.ts
```diff
@@ -62,38 +73,22 @@ export class GbVariablesComponent implements OnInit {
...
-  }
-
-  get viewMode(): VariablesViewMode {
-    return this.variableService.variablesViewMode;
-  }
-
-  set viewMode(value: VariablesViewMode) {
-    this.variableService.variablesViewMode = value;
+  get allChecked(): boolean {
+    this._allChecked = this.variableService.variables.every((v: Concept) => {
```
forus: What about having this iteration inside the service and call `this.variableService.allChecked()` instead?

src/app/modules/gb-side-panel-module/accordion-components/gb-variables/gb-variables.component.ts
beaugogh: if `listAvailableViewModes()` is used just once and this function only contains a mapper function inside, is it still a good idea ? :)

src/app/modules/gb-side-panel-module/accordion-components/gb-variables/gb-variables.component.ts
beaugogh: yes, this is debatable, my idea is to only put things inside services when it is really necessary, otherwise it is really easy to blow up services and make them unmanageable. in this case, the allChecked flag is just something that is required by the UI component to render a checkbox, it is not used anywhere else, so it is sth. more belonging to the component. The services are for storing shared variables and necessary fetching and  transformation logics on those variables.

src/app/modules/gb-side-panel-module/accordion-components/gb-variables/gb-variables.component.ts
beaugogh: since there are extra processes inside these getter and setter, yes, this _allChecked field is necessary. To bind an Angular directive model to a component field, the field itself should be static, otherwise you may experience unstable behavior on UI.

src/app/modules/gb-side-panel-module/accordion-components/gb-variables/gb-variables.component.ts
forus: I think so.
it's +5 lines in the init block.
IMHO It takes slightly less time to read this code:
```
this.availableViewModes = listAvailableViewModes()
this.viewMode = VariablesViewMode.TREE_VIEW;
```
If I'd want to know how we fetch those modes I could always jump to the function definition.

But indeed it could be a matter of taste.

PR #532, opened 2019-02-27
tmt-719 refactor fractalis workflow

beaugogh:


src/app/services/navbar.service.ts
```diff
@@ -47,11 +49,24 @@ export class NavbarService {
...
     } else if (this.isAnalysis) {
       this.activeItem = this._items[1];
+      this.configFractalis();
     } else if (this.isExport) {
       this.activeItem = this._items[2];
     }
   }
 
+  private configFractalis() {
+    if (this.firstTimeAnalysisTabActivated) {
```
gijskant: Instead of storing this here, you could ask the fractalis service if it was already setup.

src/app/services/navbar.service.ts
```diff
@@ -47,11 +48,18 @@ export class NavbarService {
...
     } else if (this.isAnalysis) {
       this.activeItem = this._items[1];
+      this.configFractalis();
     } else if (this.isExport) {
       this.activeItem = this._items[2];
     }
   }
 
+  private configFractalis() {
+    if (!this.fractalisService.F) {
```
gijskant: Seems better not to expose the `F` object, but to encapsulate it in the fractalis service. Having a public property `isInitialised` or something is better and more to the point.

src/app/services/fractalis.service.ts
```diff
@@ -64,23 +62,20 @@ export class FractalisService {
...
+    const enabled = this.appConfig.getConfig('enable-fractalis-analysis');
+    if (enabled && !fjs.fractalis) {
       MessageHelper.alert('error', 'Failed to import Fractalis.');
     } else {
-      this.setupFractalis();
-      this.selectedVariablesUpdated.asObservable().subscribe(variables => {
-        this.prepareCache(variables);
-      });
       this.timer = setInterval(() =>
         this.updateVariablesStatus(), FractalisService.TIMEOUT_VARIABLE_STATUS_UPDATE);
```
gijskant: This already starts polling the variable statuses and causes an error, because the `F` object is not yet defined.

PR #534, opened 2019-03-04
Add possibility to select dimension while building a constraint

ewelinagr:
- [TMT-723](https://jira.thehyve.nl/browse/TMT-723) - Add cohort type selection box for panels and change constraint building logic
- [TMT-725](https://jira.thehyve.nl/browse/TMT-725) - Show subject counts independently on what cohort type is selected
- [TMT-733](https://jira.thehyve.nl/browse/TMT-733) - Add optional row filter criteria for packer-export

1 issue left: 
- [x] Unneeded groupbox for the first constraint box, while restoring cohort


cypress/integration/cohort-save.feature
```diff
@@ -19,3 +19,17 @@ Feature: save cohort
     Given I am on the cohort-selection tab
     When I delete the cohort 'Query Vital Signs'
     Then the cohort 'Query Vital Signs' is deleted
+
+  Scenario: save and delete simple biomaterial level cohort
```
beaugogh: level -> dimension

cypress/integration/cohort-save.feature
```diff
@@ -19,3 +19,17 @@ Feature: save cohort
...
+    Given I am on the cohort-selection tab
+    When I create a cohort with 'Biomaterial ID' dimension constraint
+    And I save the Cohort with name 'Biomaterial ID'
+    And I restore the cohort 'Biomaterial ID'
+    Then the current cohort has biomaterial selected
+
+  Scenario: save and delete complex multi-level cohort
```
beaugogh: multi-level -> multi-dimension

cypress/integration/constraint-creation.feature
```diff
@@ -37,3 +37,8 @@ Feature: create constraints by selecting nodes from the tree
     When I select patients that are part of study Oracle_1000_Patient with age between 50 - 55  and numerical_1 between 0 - 10
     Then there are '495' subjects
 
+  Scenario: create a simple sample based subselection constraint
```
beaugogh: a little confused with the terminology, does 'sample-based' only refer to diagnosis? or it meant for all of the four: subject, diagnosis, biomaterial and bio-source?

cypress/support/step_definitions/cohort-save.js
```diff
@@ -100,10 +100,53 @@ when('I save the Cohort with name {string}', (cohortName) => {
...
+
+  cy.drag('CLINICAL_TRIAL ').drop(1);
+  cy.get('.update-btn').eq(0).click();
+
+})
+
+then('the current cohort has multiple dimensions selected', () => {
+  cy.get('.gb-constraint-cohort-type-dropdown').eq(0).contains('Biomaterial ID');
+  cy.get('.gb-constraint-cohort-type-dropdown').eq(1).contains('patient');
+  cy.get('.gb-constraint-cohort-type-dropdown').should('have.length', 2); //FIXME
```
beaugogh: 'FIXME': is it fixed? :D

cypress/support/step_definitions/constraint-creation.js
```diff
@@ -100,3 +100,13 @@ when("I use negated pedigree constraint {string} with concept {string} and negat
   cy.get('.ui-inputswitch-slider').eq(2).click();
   cy.get('.update-btn').eq(0).click();
 });
+
+when("I select diagnosis that are part of study {string}", (studyName) => {
```
beaugogh: singular: diagnosis, plural: diagnoses,
or you meant it as a mass noun, in which case, the aux verb would be 'is'

cypress/support/step_definitions/cohort-save.js
```diff
@@ -100,10 +100,53 @@ when('I save the Cohort with name {string}', (cohortName) => {
...
+  cy.get('gb-combination-constraint').should('have.length', 2); //FIXME
```
beaugogh: see the above comment

src/app/models/cohort-models/dimension.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export class Dimension {
+  private _name: string;
+  private _sortIndex: number;
+
+  constructor(label: string) {
+    this.name = label;
+  }
+
+  get sortIndex(): number {
```
beaugogh: so the idea of the models is for glowing bear to construct VISUAL elements, except for transmart-models, if certain attributes, incl. getters and setters do not serve the purpose of visual rendering, normally we move such attributes to transmart-models, and add a mapper. (but maybe such design choice has been violated too many times...)

cypress/support/step_definitions/cohort-save.js
ewelinagr: not yet, I mentioned it in the description of this PR ("2 issues left") - so this is caused by the second one

src/app/models/constraint-models/combination-constraint.ts
```diff
@@ -107,4 +106,8 @@ export class CombinationConstraint extends Constraint {
       return new TrueConstraint();
     }
   }
+
+  private get modifiesLowerLevels(): boolean {
```
beaugogh: can there be a better name for this function?

src/app/models/constraint-models/constraint.ts
```diff
@@ -10,6 +10,8 @@ import {ConstraintMark} from './constraint-mark';
 
 export class Constraint {
 
+  static readonly BASIC_DIMENSION: string = 'patient';
```
beaugogh: maybe it is 'DEFAULT_DIMENSION'?

beaugogh: shouldn't this be an enum?

src/app/services/cohort.service.ts
```diff
@@ -27,6 +27,7 @@ import {CountService} from './count.service';
 import {SubjectSetConstraint} from '../models/constraint-models/subject-set-constraint';
 import {SubjectSet} from '../models/constraint-models/subject-set';
 import {Constraint} from '../models/constraint-models/constraint';
+import {Dimension} from '../models/cohort-models/dimension';
```
beaugogh: dimension model inside cohort-models or constraint-models?

src/app/services/http/transmart-http.service.ts
```diff
@@ -142,7 +143,7 @@ export class TransmartHttpService {
    */
   getCountsPerStudyAndConcept(constraint: Constraint): Observable<object> {
     const urlPart = 'observations/counts_per_study_and_concept';
-    const body = {constraint: TransmartConstraintMapper.mapConstraint(constraint)};
+    const body = {constraint: TransmartConstraintMapper.mapConstraintOnPatientLevel(constraint)};
```
beaugogh: mapConstraint is supposed to be a generic method, why change it to a specific one?

src/app/utilities/transmart-utilities/transmart-mapper.ts
```diff
@@ -255,4 +251,13 @@ export class TransmartMapper {
     return map;
   }
 
+  public static mapDimensions(transmartDimensions: TransmartDimension[]) {
+    let cohortTypes = [];
+    transmartDimensions.forEach( (transmartDimension: TransmartDimension) => {
+      let dimension = new Dimension(transmartDimension.name);
+      dimension.sortIndex = transmartDimension.sortIndex;
```
beaugogh: why is the sortIndex assignment? what is it used for?

cypress/support/step_definitions/cohort-save.js
ewelinagr: now it is fixed

cypress/support/step_definitions/constraint-creation.js
ewelinagr: thanks, I meant diagnoses, it is changed now

src/app/services/cohort.service.ts
ewelinagr: it was previously named `cohortType`, that's why it ended up here, but indeed after renaming constraint models seems to be a better choice. I will also move the dimension fetching function to the constraint service (instead of the cohort service)

src/app/utilities/transmart-utilities/transmart-mapper.ts
ewelinagr: this is to make sure dimensions in the dropdown are shown in a proper order. It is important because of the hierarchy between dimensions (biomaterial belongs to biosource, biosource -> diagnosis, diagnosis -> patient, image -> patient)
I just realized the sorting is not working correctly now, will fix it.

src/app/services/http/transmart-http.service.ts
ewelinagr: it is because most of the constraints have to be wrapped into patient subselection (except i.e. /query calls) .` mapConstraintOnPatientLevel` is basically `mapConstraint` wrapped into patient subselection.

src/app/models/constraint-models/constraint.ts
ewelinagr: I named it `BASIC_DIMESENSION` to show that patient is on top of other dimensions in the hierarchy. It is indeed selected by default, so I am not against renaming it, if you think it suits better. Not sure about an enum, because we don't want to hardcode other dimensions, `patient` dimension is the only special one.

src/app/models/cohort-models/dimension.ts
ewelinagr: it is a bit related to a visual part, because it should determine the order of dimensions in the dropdown (see one of the comments below)

src/app/models/constraint-models/combination-constraint.ts
ewelinagr: true, I should stick to the 'dimension' naming

cypress/support/step_definitions/constraint-creation.js
beaugogh: still the same?

src/app/services/http/transmart-http.service.ts
beaugogh: so how exactly does the sortIndex work?

src/app/utilities/transmart-utilities/transmart-mapper.ts
beaugogh: how exactly does sortIndex work?

src/app/services/http/transmart-http.service.ts
beaugogh: then we can still call it mapConstraint, right

beaugogh: if the mark is 'subject', mapConstraint will wrap it with patient subselction, there is no need to hard code it

src/app/models/constraint-models/constraint.ts
beaugogh: why is patient different from the other dimensions?

src/app/models/constraint-models/combination-constraint.ts
beaugogh: also the naming for boolean getters normally follows: isFooBar()

src/app/models/cohort-models/dimension.ts
beaugogh: not sure how sortIndex works, how is it ordered? alphabetically? doesn't seem like so

ewelinagr: yes, it was not working correctly, sortIndex is removed now, sorting is fixed now and happens in the transmart-mapper

src/app/models/constraint-models/constraint.ts
ewelinagr: because GB needs to know that this is the top level dimension. It is the only one required (others are optional) and also it is used, when there is a need to fetch all the data (count calls, variable callse etc.) - wrapping in patient subselection

src/app/services/http/transmart-http.service.ts
ewelinagr: when a mark is subject, it is wrapped into selected dimension subselection, which is not always patient. For some calls it has to stay like this, but for others there is a need for this additional layer of subselection (patient based).

src/app/utilities/transmart-utilities/transmart-mapper.ts
ewelinagr: it is just a number defining the position on the dimension list

src/app/models/constraint-models/constraint.ts
beaugogh: just that naming it 'BASIC_DIMENSION' doesn't seem to imply that it is the top-level dimension, we could just name it 'TOP_LEVEL_DIMENSION'.

src/app/services/http/transmart-http.service.ts
beaugogh: if we always need to provide the patient subselction wrapper, then maybe we shouldn't use 'mark' to denote it, since you always add that wrapper. Also, maybe we should rename 'mark' to 'dimension' as it makes more sense.

This is what I mean: if the dimension of constraint `c` is 'biomaterial', then 
`c.dimension = ConstraintDimension.BIOMATERIAL`

and when you try to convert `c` to a constraint object, you will wrap it with patient subselection, in a mapper e.g. :

`public toConstraintObject(c: Constraint): object {
    return {
        type: patient-subselection,
        constraint: c.toConstraintObject()
    }
}`

does this make sense?

beaugogh: also, it seems that wrapping constraints with patient subselection on the top level is due to the technical requirements from the way transmart works, maybe sth. we can separate from how glowing bear converts constraint to objects?

src/app/modules/gb-cohort-selection-module/constraint-components/gb-combination-constraint/gb-combination-constraint.component.ts
```diff
@@ -28,8 +28,27 @@ export class GbCombinationConstraintComponent extends GbConstraintComponent impl
 
   searchResults: Constraint[];
   selectedConstraint: Constraint;
+  dimensions: SelectItem[];
+
+  private static dimensionToDimensionOption(validDimensions) {
```
gijskant: `validDimensions: Dimension[]`

src/app/services/http/transmart-packer-http.service.ts
```diff
@@ -85,14 +85,18 @@ export class TransmartPackerHttpService {
   runJob(jobName: string, targetConstraint: Constraint): Observable<TransmartPackerJob> {
     const urlPart = `jobs/create`;
     const responseField = 'job';
+    const constraint = TransmartConstraintMapper.mapConstraintOnPatientLevel(targetConstraint);
 
     let body = {
       job_type: this.customExportJobName,
       job_parameters: {
-        constraint: TransmartConstraintMapper.mapConstraint(targetConstraint),
```
gijskant: Wouldn't it be useful to specify a type for the body of this call?

src/app/services/mocks/constraint.service.mock.ts
```diff
@@ -93,4 +97,20 @@ export class ConstraintServiceMock {
...
 
+  get validDimensionsUpdated(): Subject<Cohort[]> {
+    return this._validDimensionsUpdated;
+  }
+
+  set validDimensionsUpdated(value: Subject<Cohort[]>) {
+    this._validDimensionsUpdated = value;
+  }
+
+  get validDimensions(): Dimension[] {
```
gijskant: What are valid dimensions?

src/app/utilities/transmart-utilities/transmart-constraint-mapper.ts
```diff
@@ -22,13 +22,23 @@ import {Study} from '../../models/constraint-models/study';
...
 import {TransmartConstraintSerialiser} from './transmart-constraint-serialiser';
-import {ConstraintHelper} from '../constraint-utilities/constraint-helper';
+import {ConstraintMark} from '../../models/constraint-models/constraint-mark';
 
 export class TransmartConstraintMapper {
 
   static fullSerialiser = new TransmartConstraintSerialiser(true);
   static defaultSerialiser = new TransmartConstraintSerialiser(false);
 
+  public static mapConstraintOnPatientLevel(constraint: Constraint, full?: boolean): object {
```
gijskant: I wonder in what cases this is needed.

src/app/utilities/constraint-utilities/constraint-helper.ts
```diff
@@ -192,10 +192,10 @@ export class ConstraintHelper {
 
   static mapDataTabletoObject(dataTable: DataTable): object {
     let obj = {};
-    obj['columnDimensions'] = dataTable.columnDimensions.map((dim: Dimension) => {
+    obj['columnDimensions'] = dataTable.columnDimensions.map((dim: TableDimension) => {
```
gijskant: This function is unused and I don't understand what it is doing in this file.

src/app/services/mocks/constraint.service.mock.ts
ewelinagr: all the dimensions that can be selected in the dropdown, when creating a constraint. I followed the naming used for pedigree constraint (`validPedigreeTypes`). I also thought about renaming this to `constraintDimensions` or something like this, but if you have other suggestions, please let me know.

src/app/utilities/transmart-utilities/transmart-constraint-serialiser.ts
```diff
@@ -60,23 +60,24 @@ export class TransmartConstraintSerialiser extends AbstractConstraintVisitor<obj
   /**
    * Wrap a given query object with subselection clause
    * @param {Object} queryObject
+   * @param {string} dimension - dimension for the subselection constraint
    * @returns {Object}
    */
-  static wrapWithSubselection(queryObject: object): object {
+  static wrapWithSubselection(queryObject: object, dimension: string): object {
```
gijskant: This class may be further refactored. I don't see a need for `combination.mark` in `visitCombinationConstraint`. If the combination constraint has a subjectdimension set, then all its children are wrapped in a subselection.
This can be done by, e.g., 
```groovy
return {
  type: combination.combinationState === CombinationState.And ? 'and' : 'or',
  args: combination.args.map(arg => this.visit(new SubselectionConstraint(combination.dimension, args)))
```
or something like that.

src/app/utilities/transmart-utilities/transmart-constraint-serialiser.ts
```diff
@@ -416,4 +418,11 @@ export class TransmartConstraintSerialiser extends AbstractConstraintVisitor<obj
     }
   }
 
+  private verifyMark(optConstraint) {
```
gijskant: The mark can be deprecated because of the dimension.

src/app/models/constraint-models/constraint.ts
```diff
@@ -18,12 +20,15 @@ export class Constraint {
   protected _parentConstraint: Constraint;
   // The negation flag indicating whether to add a logical negation to the constraints
   protected _negated: boolean;
+  // The dimension for which observations will be selected
+  private _dimension: string;
```
gijskant: I would expect this to be a field of `CombinationConstraint`. Why is it needed here?

src/app/models/constraint-models/constraint.ts
```diff
@@ -18,12 +20,15 @@ export class Constraint {
...
 
   constructor() {
     this.textRepresentation = '';
     this.mark = ConstraintMark.OBSERVATION;
```
gijskant: Looks to me that the `mark` can be removed. Its use is now superseded by the `dimension` attribute.

src/app/utilities/transmart-utilities/transmart-constraint-mapper.ts
ewelinagr: do you mean in general for which calls we need to wrap constraint in patient subselection or that this wrapping could be done differently ?

src/app/models/constraint-models/constraint.ts
beaugogh: we want to rename mark to dimension, not to add another dimension attribute, right?

ewelinagr:
@beaugogh @gijskant the `mark` is removed now, so please have a look again. I am still working on improving constraint optimizations (removing unneeded subselection groupboxes created when restoring constraint)

src/app/services/constraint.service.ts
```diff
@@ -180,7 +193,7 @@ export class ConstraintService {
     if (hasUnselected) {
       let result: CombinationConstraint = new CombinationConstraint();
       result.combinationState = CombinationState.Or;
-      result.mark = ConstraintMark.OBSERVATION;
+      result.dimension = Constraint.TOP_LEVEL_DIMENSION;
```
beaugogh: shouldn't the dimension attribute of a combination constraint default to TOP_LEVEL_DIMENSION? which also means removing TOP_LEVEL_DIMENSION from Constraint class

src/app/services/cohort.service.ts
```diff
@@ -130,46 +165,13 @@ export class CohortService {
...
-  }
-
   public updateCountsWithAllCohorts(): Promise<any> {
     return new Promise((resolve, reject) => {
       console.log('Updating counts from all cohorts...');
       this.isUpdatingAll = true;
       let combination: CombinationConstraint = new CombinationConstraint();
       combination.combinationState = CombinationState.Or;
-      combination.mark = ConstraintMark.SUBJECT;
+      combination.dimension = Constraint.TOP_LEVEL_DIMENSION;
```
beaugogh: see comments related this dimension assignment below

src/app/models/constraint-models/pedigree-constraint.ts
```diff
@@ -80,7 +79,7 @@ export class PedigreeConstraint extends Constraint {
   set rightHandSideConstraint(value: CombinationConstraint) {
     this._rightHandSideConstraint = value;
     this._rightHandSideConstraint.parentConstraint = this;
-    this._rightHandSideConstraint.mark = ConstraintMark.SUBJECT;
+    this._rightHandSideConstraint.dimension = Constraint.TOP_LEVEL_DIMENSION;
```
beaugogh: see related comment below

src/app/models/constraint-models/combination-constraint.ts
```diff
@@ -10,17 +10,16 @@ import {Constraint} from './constraint';
...
-
   private _children: Constraint[];
   private _combinationState: CombinationState;
   private _isRoot: boolean;
+  private _dimension: string;
 
   constructor(children?: Constraint[],
               state?: CombinationState,
-              mark?: ConstraintMark) {
+              dimension?: string) {
```
beaugogh: still think we should make this enum, to serve as a gate guarding what kinds of dimensions are allowed in a combination constraint. Enum is meant to deal with pre-determined / hardcoded scenarios.

src/app/models/constraint-models/combination-constraint.ts
```diff
@@ -31,7 +30,7 @@ export class CombinationConstraint extends Constraint {
       });
     }
     this.combinationState = state ? state : CombinationState.And;
-    this.mark = mark ? mark : ConstraintMark.OBSERVATION;
+    this.dimension = dimension ? dimension : Constraint.TOP_LEVEL_DIMENSION;
```
beaugogh: `this.dimension = dimension ? dimension : CombinationDimension.SUBJECT;`

also see related comments below

src/app/services/http/transmart-http.service.ts
```diff
@@ -155,7 +156,7 @@ export class TransmartHttpService {
    */
   getCountsPerStudy(constraint: Constraint): Observable<object> {
     const urlPart = 'observations/counts_per_study';
-    const body = {constraint: TransmartConstraintMapper.mapConstraint(constraint)};
+    const body = {constraint: TransmartConstraintMapper.mapConstraintOnPatientLevel(constraint)};
```
beaugogh: still not quite get the motivation behind this change, we replace mapConstraint with mapConstraintOnPatientLevel everywhere, this means mapConstraint no longer exists, right? If this is just a simple renaming of the function, I am ok with it, just seems that the new name does not make the purpose of the function clearer, maybe also add some comments explaining what the function actually does.

src/app/services/mocks/constraint.service.mock.ts
beaugogh: when people see `validDimensions`, they think there might be some invalid ones as well, how can I get the full list of dimensions?

src/app/services/http/transmart-http.service.ts
ewelinagr: `mapConstraint` is still used for some calls, e.g. in gb-backend-service for the /queries calls, also for the transmart-packer export. `mapConstraintOnPatientLevel` means map constraint as before, but additionally wrap it with patient subselection. Not sure how to rename it.

src/app/models/constraint-models/combination-constraint.ts
ewelinagr: Actually dimensions depend on the loaded data, so I don't thing hardcoding them would be a good option, e.g. they can be named differently depending on how they were loaded ("diagnosis", "Diagnosis ID" etc). "patient" dimension is the only one that should be always the same.

beaugogh: wait, these dimensions are used as modifiers in transmart db, how can data loaders decide the names of dimensions? shouldn't transmart db already have guidelines for that?

src/app/services/http/transmart-http.service.ts
beaugogh: I see, it's clear, then I think it is fine like this 👍 


ps: for now

src/app/models/constraint-models/combination-constraint.ts
ewelinagr: in transmart the dimension names end up in two tables - observations and modifier description table, but the name can be anything - they are not hardcoded in the transmart code.

src/app/utilities/transmart-utilities/transmart-constraint-mapper.ts
gijskant: Yes, in general. If, for instance, you would do a counts call for:
1. a biomaterial query, like all biomaterials of all twins;
2. the same query, but then wrapped in a patient selection.
The patient count will give the same result for both (the observation count may be different, though).

src/app/utilities/constraint-utilities/abstract-constraint-visitor.ts
```diff
@@ -48,6 +52,8 @@ export abstract class AbstractConstraintVisitor<T> implements ConstraintVisitor<
         return this.visitTrialVisitConstraint(<TrialVisitConstraint>constraint);
       case 'TimeConstraint':
         return this.visitTimeConstraint(<TimeConstraint>constraint);
+      case 'SubselectionConstraint':
```
gijskant: Am I correct that this type is used nowhere in Glowing Bear?

src/app/utilities/transmart-utilities/transmart-constraint-mapper.ts
```diff
@@ -22,13 +22,24 @@ import {Study} from '../../models/constraint-models/study';
...
+    let patientLevelConstraint: Constraint;
+    if (constraint.className === 'CombinationConstraint'
+      && (<CombinationConstraint>constraint).dimension === 'patient' || constraint.className === 'TrueConstraint') {
+      patientLevelConstraint = constraint;
+    } else {
+      patientLevelConstraint = new SubselectionConstraint('patient', constraint)
```
gijskant: Hm, this only seems to be relevant for the constraint mapper, so why not do this logic in the mapper, instead of introducing a new type here, that is only used for the mapper?

src/app/models/constraint-models/combination-constraint.ts
beaugogh: ok, I see

src/app/utilities/transmart-utilities/transmart-constraint-mapper.ts
ewelinagr: it is also used for the combination constraint visitor, as you suggested here: https://github.com/thehyve/glowing-bear/pull/534#discussion_r262399883

PR #536, opened 2019-03-06
Embed material design icons, instead of fetching from Google CDN.

gijskant:


src/styles/styles.css
```diff
@@ -3,8 +3,7 @@
 @import "../../node_modules/roboto-fontface/css/roboto/roboto-fontface.css";
 @import "../../node_modules/primeicons/primeicons.css";
 @import "./primeng-theme.css";
-/*@import "~@angular/material/prebuilt-themes/indigo-pink.css";*/
-@import "https://fonts.googleapis.com/icon?family=Material+Icons";
+@import "../../node_modules/material-design-icons/iconfont/material-icons.css";
```
beaugogh: any particular reason for this? would increase loading overhead

gijskant: The overhead is minimal and it makes sure that GB also works in environments where users do not have access to the public internet.

PR #538, opened 2019-03-08
Add cohort type and dimension icons

ewelinagr:
[TMT-729](https://jira.thehyve.nl/browse/TMT-729) - Extend cohort objects in cohort panel by adding icon and type

### DIFF
--- a/src/app/modules/gb-cohort-selection-module/constraint-components/gb-combination-constraint/gb-combination-constraint.component.ts
+++ b/src/app/modules/gb-cohort-selection-module/constraint-components/gb-combination-constraint/gb-combination-constraint.component.ts
@@ -16,6 +16,7 @@ import {PedigreeConstraint} from '../../../../models/constraint-models/pedigree-
 import {SelectItem, TreeNode} from 'primeng/api';
 import {UIHelper} from '../../../../utilities/ui-helper';
 import {Dimension} from '../../../../models/constraint-models/dimension';
+import {IconHelper} from '../../../../utilities/icon-helper';
 
 @Component({
   selector: 'gb-combination-constraint',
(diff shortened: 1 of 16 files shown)


src/app/modules/gb-cohort-selection-module/constraint-components/gb-combination-constraint/gb-combination-constraint.component.ts
```diff
@@ -155,4 +155,8 @@ export class GbCombinationConstraintComponent extends GbConstraintComponent impl
     (<CombinationConstraint>this.constraint).dimension = this.selectedDimension;
   }
 
+  getDimensionIcon(dimension: string): string {
+    return this.constraintService.getDimensionIcon(dimension);
```
gijskant: I don't understand why there are so many service calls in the PR for fecthing an icon class. Why not create one static helper function in a IconHelper class or something like that?

src/app/services/transmart-resource.service.ts
```diff
@@ -520,4 +520,30 @@ export class TransmartResourceService {
       }));
   }
 
+  /**
+   * Gets an icon per subject dimension name
+   * TODO TMT-773 - make dimension icons configurable
+   * @param dimensionName
+   */
+  getSubjectDimensionIcon(dimensionName: string): string {
```
gijskant: This does not seem to be the right place. Please create a separat class for this.

src/app/modules/gb-cohort-selection-module/constraint-components/gb-combination-constraint/gb-combination-constraint.component.ts
ewelinagr: done!

PR #539, opened 2019-03-11
Refactor Transmart constraint mapper and cohort mapper.

gijskant:
The main purpose of these changes is to make serialisation of constraints to transmart and parsing imported constraints more structured and more tested. I chose to use the visitor pattern again for parsing constraints (TransmartConstraintReader), just as in serialising (TransmartConstraintSerialiser).
I also moved the cohort mappers to a separate class.
I added tests for the reader and serialiser that for each constraint type:
1. first serialise the constraint
2. then parse the constraint
3. then serialise the parsed constraint
4. check if the results of 1 and 3 are equal.

This does not cover all that we want to test, but at least it tests: if I save and restore a query, does it then still give the same results?
It does not cover the translation of all constraint types to the correct serialisation for Transmart.

Tasks left:
- TODO: refactor `visitCombinationConstraint` in [transmart-constraint-reader.ts](/thehyve/glowing-bear/tree/refactor-transmart-constraint-mapper/src/app/utilities/transmart-utilities/transmart-constraint-reader.ts#L76)
- TODO: discuss serialisation of dimension subselection by some examples
- Unit tests added for:
  - [x] transmart-constraint-serialiser
  - [x] transmart-constraint-reader
  - [x] transmart-cohort-mapper

And did you notice the >80% coverage?


src/app/models/transmart-models/transmart-cohort.ts
```diff
@@ -0,0 +1,23 @@
+import {CohortSubscriptionFrequency} from '../cohort-models/cohort-subscription-frequency';
+import {TransmartConstraint} from './transmart-constraint';
+
+export class TransmartCohort {
```
ewelinagr: Looks like a duplication of [GbBackendQuery](src/app/models/gb-backend-models/gb-backend-query.ts) class

gijskant: Ha, interesting. Wonder why there were two mappers in the first place then. I will merge them.

src/app/utilities/transmart-utilities/transmart-constraint-serialiser.ts
```diff
@@ -112,27 +150,24 @@ export class TransmartConstraintSerialiser extends AbstractConstraintVisitor<obj
...
+    const result: TransmartTimeConstraint = {
       type: 'time',
       field: {
         dimension: dimension,
         fieldName: fieldName,
-        type: 'DATE'
+        type: TransmartType.date
       },
-      operator: operator,
+      operator: <TransmartOperator>operator,
```
ewelinagr: Looks like there is some miss-match between `DateOperatorState` symbols and `TransmartOperator` (before is '<=' and '<-' accordingly):
https://github.com/thehyve/glowing-bear/pull/539/files#diff-7ff27abe4af0fc943ddd37e4dcd068adR138

src/app/models/transmart-models/transmart-constraint.ts
```diff
@@ -0,0 +1,136 @@
...
+  conceptCodes?: string[];
+  path?: string;
+}
+
+export class TransmartStudyNameConstraint extends TransmartConstraint {
+  type = 'study_name';
+  studyId: string;
+}
+
+export class TransmartNullConstraint extends TransmartConstraint {
```
ewelinagr: According to swagger Null constraint has also required `field` property

src/app/utilities/transmart-utilities/transmart-cohort-mapper.ts
```diff
@@ -0,0 +1,55 @@
+import {Cohort} from '../../models/cohort-models/cohort';
+import {TransmartConstraintMapper} from './transmart-constraint-mapper';
+import {TransmartCohort} from '../../models/transmart-models/transmart-cohort';
+
+export class TransmartCohortMapper {
```
ewelinagr: this should be also merged with [GbBackendMapper](src/app/utilities/gb-backend-mapper.ts), right?

gijskant: indeed

ewelinagr:
I tested a bit saving and restoring of cohorts, it seems that there are some issues with that, e.g. when creating a query: "give me all biomaterials that are linked to a patients that have observations for concept gender" I get proper counts, but when I save it and restore the patient subselection inside biomaterial subselection is gone and I have 0 results.

src/app/models/transmart-models/transmart-cohort.ts
gijskant: There is an interesting difference:
In `GbBackendQuery`:
- `queryConstraint` (`object`)
- `queryBlob` (`object`)
In `TransmartCohort`:
- `constraint` (`TransmartConstraint`)

gijskant:
Thanks for the test case, I will investigate. The other issues should be resolved now.

gijskant:
The restoring should be fixed now as well, by (de)serialising the queryBlob.

ewelinagr:
There are some more cases I found that still seem not to be working correctly:
case 1:
saved:
sav1
restored:
res1

case 2:
saved:
sav2
restored:
res2

src/app/utilities/constraint-utilities/constraint-copier.ts
```diff
@@ -0,0 +1,111 @@
...
+import {NegationConstraint} from '../../models/constraint-models/negation-constraint';
+import {ConceptConstraint} from '../../models/constraint-models/concept-constraint';
+import {CombinationConstraint} from '../../models/constraint-models/combination-constraint';
+import {Constraint} from '../../models/constraint-models/constraint';
+
+/**
+ * Constraint copy class, for creating a copy of a constraint that can be
+ * edited independently.
+ */
+export class ConstraintCopier extends AbstractConstraintVisitor<Constraint> {
```
beaugogh: This could be done by polymorphism for the Constraint class and its descendants, e.g. a method named `copy()`, so that the copying is done generically and one does not need to call these different visitor functions for different types of constraints

src/app/services/http/gb-backend-http.service.ts
```diff
@@ -40,18 +40,18 @@ export class GbBackendHttpService {
    * Get the queries that the current user has saved.
    * @returns {Observable<Query[]>}
    */
-  getQueries(): Observable<GbBackendQuery[]> {
+  getQueries(): Observable<CohortRepresentation[]> {
```
beaugogh: maybe the naming could be getCohorts

beaugogh: btw, the name 'CohortRepresentation' sounds odd, not sure how to improve it though.

src/app/services/http/transmart-http.service.ts
```diff
@@ -143,7 +142,7 @@ export class TransmartHttpService {
    */
   getCountsPerStudyAndConcept(constraint: Constraint): Observable<object> {
     const urlPart = 'observations/counts_per_study_and_concept';
-    const body = {constraint: TransmartConstraintMapper.mapConstraintOnPatientLevel(constraint)};
+    const body = {constraint: TransmartConstraintMapper.mapConstraint(constraint)};
```
beaugogh: looks good

src/app/utilities/transmart-utilities/transmart-constraint-reader.ts
```diff
@@ -0,0 +1,321 @@
...
+import {DateOperatorState} from '../../models/constraint-models/date-operator-state';
+import {Operator} from '../../models/constraint-models/operator';
+
+/**
+ * Deserialisation class for reading constraint objects from TranSMART API and
+ * transform them into constraints that can be used in the cohort selection.
+ *
+ * To deserialise a constraint, use <code>new TransmartConstraintReader().visit(constraint)</code>.
+ */
+export class TransmartConstraintReader extends AbstractTransmartConstraintVisitor<Constraint> {
```
beaugogh: as I understand, CohortMapper does both serialisation and deserialisation for cohorts, one would expect there is also a TransmartConstraintMapper that does serialisation and deserialisation for transmart constraints, right?

gijskant: There are two reasons for this: 
1. the transmart constraint mappers implement the visitor interfaces
2. size

src/app/utilities/constraint-utilities/constraint-copier.ts
gijskant: Yes, indeed, that is another approach that would work. It would make the constraint classes a bit heavy in my opinion. E.g., also serialisation could be approached that way, but then you have:
- the serialisation/copying code scattered over several classes
- the constraint classes carrying many responsibilities (capture properties, cloning, serlalisation, ...)

beaugogh:
generally I like the refactoring! just some minor comments and close the loop holes as @ewelinagr  pointed out

gijskant:
@ewelinagr The reported cases should be fixed now. Could you have a look again?

src/app/utilities/constraint-utilities/constraint-copier.ts
beaugogh: serialisation indeed needs to be separated as it is about converting glowing bear constraints to a backend-specific format, whereas the copy() method is another story.

src/app/models/constraint-models/constraint.clone.spec.ts
```diff
@@ -1,30 +1,28 @@
...
+import {StudyConstraint} from './study-constraint';
+import {PedigreeConstraint} from './pedigree-constraint';
+import {CombinationConstraint} from './combination-constraint';
+import {ConceptConstraint} from './concept-constraint';
+import {TrialVisit} from './trial-visit';
+import {Study} from './study';
+import {Concept} from './concept';
 
-describe('ConstraintCopier', () => {
+describe('Constraint.clone', () => {
```
ewelinagr: couldn't it be a part of constraint.spec class?

gijskant: Has been moved now.

ewelinagr:
A part of this change is also that the study constraint is not auto-added, when dragging a concept node, right? Looks like it broke some cypress tests (constraint-creation.feature )

src/app/utilities/constraint-utilities/constraint-copier.ts
gijskant: Moved to the Constraint classes.

gijskant:
Ah, yes indeed. That is a side effect of fixing the behaviour. What was happening before was not correct. The correct behaviour will be added in a separate branch, implementing [TMT-770](https://jira.thehyve.nl/browse/TMT-770).

src/app/utilities/transmart-utilities/transmart-constraint-reader.ts
```diff
@@ -0,0 +1,321 @@
...
+    if (rightHandSide.className === 'CombinationConstraint') {
+      const combiRhs =  <CombinationConstraint>rightHandSide;
+      for (let child of combiRhs.children) {
+        constraint.rightHandSideConstraint.addChild(child);
+      }
+      constraint.rightHandSideConstraint.combinationState = combiRhs.combinationState;
+    } else if (rightHandSide.className !== 'TrueConstraint') {
+      constraint.rightHandSideConstraint.addChild(rightHandSide);
+    } else {
+      // FIXME: no restoring of other constraint types?
```
ewelinagr: is this something to be fixed in this PR?

ewelinagr:
Looks good, great improvement and really nice test coverage! The issues with the test cases I described earlier seem to be fixed now. 
The only problem left is with wrapping the constraints in the patient subselection for some calls - this PR will break [TMT-725](https://jira.thehyve.nl/browse/TMT-725) and the special packer export ([TMT-733](https://jira.thehyve.nl/browse/TMT-733)

PR #540, opened 2019-03-11
Add user-friendly constraint box messages

ewelinagr:
[TMT-727](https://jira.thehyve.nl/browse/TMT-727)

### DIFF
--- a/src/app/services/constraint.service.ts
+++ b/src/app/services/constraint.service.ts
@@ -290,6 +290,17 @@ export class ConstraintService {
     return depth;
   }
 
+  public parentDimension(constraint: Constraint): string {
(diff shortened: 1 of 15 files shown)


src/app/services/constraint.service.ts
```diff
@@ -290,6 +290,17 @@ export class ConstraintService {
     return depth;
   }
 
+  public parentDimension(constraint: Constraint): string {
```
gijskant: Maybe move this to the Constraint class?

ewelinagr:
@gijskant I moved `parentDimension` as you suggested, together with `constraintDepth` (similar reason that you described).

gijskant:
:+1:

PR #541, opened 2019-03-12
Update cypress tests according to the latest changes

ewelinagr:
Remaining failing test is related to a back-end issue (see [TMT-778](https://jira.thehyve.nl/browse/TMT-778))

### DIFF
--- a/cypress/support/step_definitions/constraint-creation.js
+++ b/cypress/support/step_definitions/constraint-creation.js
@@ -70,8 +70,8 @@ when("I select patients that are part of study Oracle_1000_Patient with age betw
   cy.get('input[placeholder="max:65"]').type('55');
   cy.contains('numerical_1 ').trigger('dragstart');
   cy.get('input[placeholder="add criterion"]').eq(1).trigger('drop');
-  cy.get('input[placeholder="min:-5.76679"]').type('0');
-  cy.get('input[placeholder="max:29.79301"]').type('10');
+  cy.get('input[placeholder="min:5.766787251"]').type('0');
+  cy.get('input[placeholder="max:29.7930135202"]').type('10');
   cy.get('.update-btn').eq(0).click();
 });
 
--- a/cypress/support/step_definitions/fractalis-charts.js
+++ b/cypress/support/step_definitions/fractalis-charts.js
@@ -17,21 +17,16 @@ then('There should be a box plot created', () => {
 
 then('I select two cohorts', () => {
   cy.get('span.ui-button-text.ui-unselectable-text').then(cohortsElms => {
-    if(cohortsElms && cohortsElms.length > 1) {
-      // if there is already more than one cohort, select the 2nd one
-      cohortsElms.eq(1).click();
-    } else {
-      // else creat and save a new cohort, then select it
-      const cohortName = 'temp_oracle_numerical';
-      cy.toggleNode('Public Studies');
-      cy.toggleNode('Oracle_1000_Patient');
-      cy.toggleNode('Numerical Variables');
-      cy.drag('numerical_11').drop(0);
-      cy.contains('Update cohort').click();
-      cy.get('#cohortName').type(cohortName);
-      cy.contains('Save cohort').click();
-      cy.get('gb-cohorts').contains(cohortName).eq(0).click();
-    }
+    // create a new cohort
+    const cohortName = 'temp_oracle_numerical';
+    cy.toggleNode('Public Studies');
+    cy.toggleNode('Oracle_1000_Patient');
+    cy.toggleNode('Numerical Variables');
+    cy.drag('numerical_11').drop(0);
+    cy.contains('Update cohort').click();
+    cy.get('#cohortName').type(cohortName);
+    cy.contains('Save cohort').click();
+    cy.get('gb-cohorts').contains(cohortName).eq(0).click();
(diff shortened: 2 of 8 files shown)


cypress/support/step_definitions/constraint-creation.js
```diff
@@ -70,8 +70,8 @@ when("I select patients that are part of study Oracle_1000_Patient with age betw
   cy.get('input[placeholder="max:65"]').type('55');
   cy.contains('numerical_1 ').trigger('dragstart');
   cy.get('input[placeholder="add criterion"]').eq(1).trigger('drop');
-  cy.get('input[placeholder="min:-5.76679"]').type('0');
-  cy.get('input[placeholder="max:29.79301"]').type('10');
+  cy.get('input[placeholder="min:5.766787251"]').type('0');
```
beaugogh: test failing

beaugogh: another failing test: 

> create a constraint with concept negation

beaugogh: looks like transmart-dev not returning the right response

cypress/support/step_definitions/fractalis-charts.js
```diff
@@ -17,21 +17,16 @@ then('There should be a box plot created', () => {
...
-      cy.toggleNode('Public Studies');
-      cy.toggleNode('Oracle_1000_Patient');
-      cy.toggleNode('Numerical Variables');
-      cy.drag('numerical_11').drop(0);
-      cy.contains('Update cohort').click();
-      cy.get('#cohortName').type(cohortName);
-      cy.contains('Save cohort').click();
-      cy.get('gb-cohorts').contains(cohortName).eq(0).click();
-    }
+    // create a new cohort
```
beaugogh: why changing this logic?

beaugogh: this test is failing:

> Create a box-plot with a numerical variable from the Category View with multiple cohorts
 expected <g.fjs-box> to have a length of 2 but got 1

cypress/support/step_definitions/constraint-creation.js
ewelinagr: yes, this is what I mentioned in the description of this PR -> back-end issue addressed in TMT-778

cypress/support/step_definitions/fractalis-charts.js
ewelinagr: hmm, strange, it's passing for me locally. 
I changed this logic, because it was selecting any cohort that was in the cohort panel, which does not have to be the expected one - with so many tests and when testing on the dev env there is a high chance there will be more cohorts. Creating it every time, when the test is run makes this test more independent.

beaugogh: > there is a high chance there will be more cohorts.

This is what we want, right?

ewelinagr: yes, but the test expects to find 'numerical_11' variable in the cohort, which is not always the case (that is why it is failing currently).

beaugogh: ?? the test expects to see two boxplots regardless which cohorts are selected, as long as there are two cohorts, where is the numerical_11 part?

ewelinagr: https://github.com/thehyve/glowing-bear/blob/625d4bb975ab75aa62edf836c4b6f1c674e0bb7b/cypress/integration/fractalis-charts.feature#L14

beaugogh: ah, I see, you are right, there is the variable dependency

beaugogh: btw, the test is still failing:

> Create a box-plot with a numerical variable from the Category View with multiple cohorts
expected <g.fjs-box> to have a length of 2 but got 1

ewelinagr: local gb with transmart-dev and gb-backend-dev. Are you for sure up to date with the recent change from this branch? There is one commit that was added after the PR was created

beaugogh: trying run with default cypress config?

> {
  "baseUrl": "https://glowingbear-dev.thehyve.net",
  "chromeWebSecurity": false,
  "video": false,
  "fixturesFolder": "cypress/fixtures/dev",
  "env": {
    "apiUrl": "https://transmart-dev.thehyve.net",
    "authentication-service-type": "oidc",
    "oidc-server-url": "https://keycloak-dwh-test.thehyve.net/auth/realms/transmart-dev/protocol/openid-connect",
    "oidc-client-id": "transmart-client",
    "gb-backend-url": "https://gb-backend-dev.thehyve.net"
  },
  "numTestsKeptInMemory": 3,
  "defaultCommandTimeout": 30000
}

ewelinagr: ah yes, but to fix this test there was a change added to the gb code (https://github.com/thehyve/glowing-bear/pull/541/commits/625d4bb975ab75aa62edf836c4b6f1c674e0bb7b), so the fix is only in this branch, "https://glowingbear-dev.thehyve.net" does not have

beaugogh: right, I forgot about this part, thanks

PR #544, opened 2019-03-14
 Add concept to dimensions mapping and dimension validation

ewelinagr:
[TMT-745](https://jira.thehyve.nl/browse/TMT-745) - mapping of concept-to-dimension
[TMT-747](https://jira.thehyve.nl/browse/TMT-747) - box validation: add auto-wrapping of concepts into dimension box if required
[TMT-748](https://jira.thehyve.nl/browse/TMT-748) - box validation: pedigree constraints - restricted to the patient dimension

### DIFF
--- a/src/app/services/tree-node.service.ts
+++ b/src/app/services/tree-node.service.ts
@@ -257,6 +257,9 @@ export class TreeNodeService {
       concept.code = treeNode.conceptCode;
       concept.fullName = treeNode.fullName;
       concept.name = treeNode.name;
+      if (treeNode.metadata && treeNode.metadata['subject_dimension']) {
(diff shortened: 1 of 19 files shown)


src/app/services/tree-node.service.ts
```diff
@@ -257,6 +257,9 @@ export class TreeNodeService {
       concept.code = treeNode.conceptCode;
       concept.fullName = treeNode.fullName;
       concept.name = treeNode.name;
+      if (treeNode.metadata && treeNode.metadata['subject_dimension']) {
```
gijskant: I was wondering why we would restrict to a single subject dimension, but since the metadata is a simple key-value map, this is a good approach. If we want to support multiple dimensions later, we could parse a `subject_dimensions` metadata element and choose a proper separator character then.

PR #545, opened 2019-03-14
Add observation level study constraints.

gijskant:
Fixes [TMT-770](https://jira.thehyve.nl/browse/TMT-770).

### DIFF
--- a/src/app/modules/gb-cohort-selection-module/constraint-components/gb-concept-constraint/gb-concept-constraint.component.ts
+++ b/src/app/modules/gb-cohort-selection-module/constraint-components/gb-concept-constraint/gb-concept-constraint.component.ts
@@ -666,6 +666,9 @@ export class GbConceptConstraintComponent extends GbConstraintComponent implemen
       !currentStudyIds.includes(study.id)
     );
     newStudies.forEach(study => conceptConstraint.studyConstraint.studies.push(study));
+    if (conceptConstraint.studyConstraint.studies.length > 0) {
+      conceptConstraint.applyStudyConstraint = true;
+    }
   }
 
   onDrop(event: DragEvent) {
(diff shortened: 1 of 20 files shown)


src/app/models/constraint-models/combination-constraint.ts
```diff
@@ -101,7 +101,7 @@ export class CombinationConstraint extends Constraint {
 
   clone(): CombinationConstraint {
     const clone = new CombinationConstraint(
-      this.children.map(child => child.clone()),
+      this.children.filter(child => child).map(child => child.clone()),
```
ewelinagr: why is this filter added?

gijskant: Sometimes I was getting `null` children. I don't understand why, actually. I will try what happens if I remove it.

src/app/modules/gb-cohort-selection-module/constraint-components/gb-concept-constraint/gb-concept-constraint.component.ts
```diff
@@ -632,6 +653,16 @@ export class GbConceptConstraintComponent extends GbConstraintComponent implemen
...
 
+  onDragEnter(event) {
+    event.stopPropagation();
+    event.preventDefault();
+  }
+
+  onDragOver(event) {
+    event.stopPropagation();
+    event.preventDefault();
+  }
```
ewelinagr: Why these functions has to be overwritten here? They are already defined like this in the parent class (GbConstraintComponent)

gijskant: Dragging behaviour has now been fixed:
- dragging study constraints to concept or study constraints, adds the studies to the existing constraint
- otherwise, constraints are added to the parent combination constraint.

PR #546, opened 2019-03-15
Reuse subject set number for counts when possible

forus:


src/app/services/count.service.spec.ts
```diff
@@ -43,4 +44,39 @@ describe('CountService', () => {
         done();
       });
   });
+
+  it('should request counts if non subject set aware constraint', () => {
+    let resourceServiceCounts = spyOn(countService['resourceService'], 'getCounts');
```
gijskant: You can also specify `resourceService = TestBed.get(ResourceService);` in the `beforeEach` block, just as for the count service, and `spyOn(resourceService, 'getCounts')`.

src/app/services/count.service.spec.ts
```diff
@@ -43,4 +44,39 @@ describe('CountService', () => {
...
+    countService.showObservationCounts = false;
+    let subjectSetConstraint = new SubjectSetConstraint();
+    subjectSetConstraint.setSize = 10;
+
+    countService.getCounts(subjectSetConstraint).subscribe((counts) => {
+      expect(counts.subjectCount).toEqual(10);
+      expect(counts.observationCount).toEqual(-1);
+    });
+
+    expect(resourceServiceCounts).toHaveBeenCalledTimes(0);
```
gijskant: Or, alternatively, `expect(resourceServiceCounts).not.toHaveBeenCalled();`.

gijskant:
Locally, you can run `yarn lint` to perform the linting check that is currently failing. The linter rules enforce one type of quotes using in imports: the single quote. We should probably edit the `.editorconfig` file to reflect this rule.

src/app/services/count.service.ts
```diff
@@ -116,6 +115,21 @@ export class CountService {
     });
   }
 
+  /**
+   * When no observation counts requested, takes advantage of subject set constraint knowing number of subjects.
+   * Falls back to making reqular call otherwise
+   * @param constraint
+   */
+  getCounts(constraint: Constraint) {
```
ewelinagr: I am still thinking whether it is better to keep this logic here or move it to transmart-resource.service, but maybe for now it is fine like this and later we will find some better place for all the optimization-related logic.

forus:
@ewelinagr @gijskant Thanks, I've just pushed the changes.

PR #548, opened 2019-03-18
Fix serialisation of negation constraints 

ewelinagr:
Fixes [TMT-778](https://jira.thehyve.nl/browse/TMT-778)


ewelinagr:
Covered also by ["create a constraint with concept negation" cypress test](https://github.com/thehyve/glowing-bear/blob/dev/cypress/integration/constraint-creation.feature#L9)

PR #549, opened 2019-03-18
Update documentation/application links in the header.

gijskant:


src/app/modules/gb-side-panel-module/gb-side-panel.component.html
```diff
@@ -1,7 +1,9 @@
 <div class="side-panel-container">
   <div class="gb-logo-panel">
-    <img src="../../../assets/icons/glowingbear.png" (click)="goToDocUrl()">
-    <span class="gb-logo-text" (click)="goToDocUrl()">GLOWING BEAR</span>
+    <a [routerLink]="''">
```
forus: Shouldn't it be `[routerLink]="['/']"` instead?

forus: I assume empty string means current path?

gijskant: It refers to the route defined here: https://github.com/thehyve/glowing-bear/blob/dev/src/app/app.routing.ts#L15.

forus: ah, ok

PR #550, opened 2019-03-19
Update dependencies.

gijskant:
Update some dependencies, mainly angular-cli, angular-split and typescript.
Also fixed some issues in MS Edge: apparently it can not iterate over a [HTMLCollection](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/children), so we need to explicitely `Array.from` to traverse DOM nodes.
Fixes [TMT-783](https://jira.thehyve.nl/browse/TMT-783).


src/app/modules/gb-cohort-selection-module/constraint-components/gb-combination-constraint/gb-combination-constraint.component.html
```diff
@@ -5,7 +5,7 @@
     <p-dropdown [options]="dimensions" [(ngModel)]="selectedDimension"
                 [disabled]="dimensions.length === 1"
                 [dropdownIcon]="dimensions.length === 1 ? '' : 'pi pi-chevron-down'"
-                (onChange)="onSelectedDimensionChange()" [autoWidth]="false" [style]="{'min-width':'10em'} ">
+                (onChange)="onSelectedDimensionChange()" [style]="{'min-width':'10em'} ">
```
beaugogh: was this in conflict with the version updates?

gijskant: indeed, the newer primeng does not provide this attribute.

gijskant: But let me check again, because only the patch version changed.

gijskant: Just checked and this change is indeed not required. Probably leftover from when I tried to upgrade primeng.

PR #553, opened 2019-03-26
Do not require fractalis url if fractalis is not enabled

ewelinagr:
Fixes [TMT-792](https://jira.thehyve.nl/browse/TMT-792)

### DIFF
--- a/src/app/config/app.config.ts
+++ b/src/app/config/app.config.ts
@@ -69,7 +69,10 @@ export class AppConfig {
           throw Error('Gb-backend URL is unspecified in the configuration.')
         }
         case 'fractalis-url': {
-          throw Error('Fractalis URL is unspecified in the configuration.')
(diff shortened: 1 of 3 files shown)


src/app/config/app.config.ts
```diff
@@ -69,7 +69,10 @@ export class AppConfig {
           throw Error('Gb-backend URL is unspecified in the configuration.')
         }
         case 'fractalis-url': {
-          throw Error('Fractalis URL is unspecified in the configuration.')
+          if (this.getConfig('enable-fractalis-analysis') === true) {
+            throw Error('Fractalis URL is unspecified in the configuration.')
+          }
+          break;
```
gijskant: the control flow is a bit confusing here... in this case, break means returning null at the very end, right? we could as well return null here.

PR #555, opened 2019-03-27
tmt-728: disable observation counts in production

beaugogh:


beaugogh:
> I think it should be changed in the puppet module (https://github.com/thehyve/puppet-glowing_bear). The default there should be false, currently it's true

the `glowing_bear::show_observation_counts` is set to false in hiera on master for glowingbear-test, should override the default value.

I also changed the default value here: https://github.com/thehyve/puppet-glowing_bear/blob/master/manifests/params.pp

PR #558, opened 2019-03-29
Fix cypress tests failing after changes for variables panel

ewelinagr:


cypress/support/step_definitions/variable-import.js
```diff
@@ -1,3 +1,8 @@
+given('I unselect all variables', () => {
+  cy.get('.checkAllText').contains('194');
```
beaugogh: something more generic maybe? 194 might change frequently

ewelinagr: hmm, true, will change it!

PR #559, opened 2019-03-29
update variable selection count

beaugogh:


src/app/services/variable.service.ts
```diff
@@ -373,7 +372,17 @@ export class VariableService {
...
-  get selectedVariablesNumber(): number {
-    return this.selectedVariablesTree.filter(node => this.treeNodeService.isTreeNodeConcept(node)).length;
+  private updateNumberOfSelectedVariables(): number {
+    let num = 0;
+    this.variables.forEach((variable: Concept) => {
+      if (variable.selected) {
+        num++;
+      }
+    });
+    return num;
```
ewelinagr: this could be also simplified, something like this? 
`this.variables.filter((variable: Concept) => {variable.selected)).length`

PR #564, opened 2019-04-09
Fix variables update flow

ewelinagr:
Fixes [TMT-815](https://jira.thehyve.nl/browse/TMT-815)

### DIFF
--- a/src/app/services/variable.service.ts
+++ b/src/app/services/variable.service.ts
@@ -67,11 +67,8 @@ export class VariableService {
     Observable.combineLatest(
       this.countService.selectedConceptCountMapUpdated.asObservable(),
       this.treeNodeService.treeNodesUpdated.asObservable()
-    ).subscribe(res => {
-      const isTreeLoadingFinished = res[1];
-      if (isTreeLoadingFinished) {
-        this.updateVariables();
-      }
+    ).subscribe(() => {
+      this.updateVariables();
     });
   }
 
--- a/src/app/services/variable.service.spec.ts
+++ b/src/app/services/variable.service.spec.ts
@@ -86,18 +86,33 @@ describe('VariableService', () => {
     expect(element).toEqual(jasmine.any(Concept));
   });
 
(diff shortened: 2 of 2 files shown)


src/app/services/variable.service.ts
```diff
@@ -67,11 +67,8 @@ export class VariableService {
     Observable.combineLatest(
       this.countService.selectedConceptCountMapUpdated.asObservable(),
       this.treeNodeService.treeNodesUpdated.asObservable()
-    ).subscribe(res => {
-      const isTreeLoadingFinished = res[1];
-      if (isTreeLoadingFinished) {
-        this.updateVariables();
-      }
```
beaugogh: should wait for tree finishing loading before updating the variables, right?

src/app/services/variable.service.ts
```diff
@@ -67,11 +67,8 @@ export class VariableService {
     Observable.combineLatest(
       this.countService.selectedConceptCountMapUpdated.asObservable(),
       this.treeNodeService.treeNodesUpdated.asObservable()
-    ).subscribe(res => {
-      const isTreeLoadingFinished = res[1];
-      if (isTreeLoadingFinished) {
```
ewelinagr: I don't understand why this condition was needed, also the `res[1]` from  `combineLatest` subscription was not always a tree ( depending on what was emitted as the latest, it was sometimes a result of the count map update)

src/app/services/variable.service.ts
ewelinagr: but the res[1] from combineLatest subscription was not always a tree ( depending on what was emitted as the latest, it was sometimes a result of the count map update). Also it looks like waiting for the tree nodes to be loaded is already a part of `updateVariables`

beaugogh: res[1] should return a boolean value indicating if the tree is finished loading, but you are right, `updateVariables` already takes care of the tree loading waiting, so we can remove the res[1] checking :)

PR #569, opened 2019-04-15
Update readme

ewelinagr:


ewelinagr:
@gijskant @beaugogh should we mention Fractalis in the readme? Some link to the repos maybe? If so - our forked ones? Or just a general fractalis documentation?

beaugogh:
> @gijskant @beaugogh should we mention Fractalis in the readme? Some link to the repos maybe? If so - our forked ones? Or just a general fractalis documentation?

I think the forked one since we use it, but the fork one will contain the reference to the original fractalis page

PR #570, opened 2019-04-16
[TMT-826] Fix cohort import

ewelinagr:


src/app/services/mocks/count.service.mock.ts
```diff
@@ -63,7 +63,13 @@ export class CountServiceMock {
     });
   }
 
-  updateAllCounts(constraint: Constraint) {
+  updateAllCounts(constraint: Constraint): Promise<any>  {
+    return new Promise<any>(resolve => {
```
gijskant: Isn't this the same as `Promise.resolve(true)`?

ewelinagr: Yes, looks like both work fine, we just use `return new Promise<any>(resolve => {` everywhere in the tests. Is it useful to change it in all the places?

gijskant: It's shorter, but indeed otherwise the same.

PR #571, opened 2019-04-16
Fix constraint mapping for packer export

ewelinagr:


src/app/services/export.service.ts
```diff
@@ -167,7 +167,8 @@ export class ExportService {
       this.resourceService.runExportJob(
         job,
         this.exportDataTypes,
-        this.variableService.combination,
+        this.constraintService.cohortSelectionConstraint,
+        this.constraintService.variableConstraint(this.variableService.variables),
```
beaugogh: this would produce a combination constraint with all the variables, is that intended?

beaugogh: what happens to the previous approach where variableService.combination is used?

src/app/utilities/transmart-utilities/transmart-constraint-serialiser.ts
```diff
@@ -354,11 +354,12 @@ export class TransmartConstraintSerialiser extends AbstractConstraintVisitor<Tra
...
     let args = children.map(child => {
-      if (parentDimension === dimension && child.type === 'subselection') {
+      if (child.type === 'subselection' && parentDimension === dimension) {
         // Subselection not required if the dimension equals the parent dimension and the child is a subselection
         return child;
-      } else if (dimension === 'observation') {
-        return child;
+      } else if (child.type === 'subselection' && (<TransmartSubSelectionConstraint>child).dimension === 'observation') {
+        // Observation-level constraint should be unwrapped
+        return (<TransmartSubSelectionConstraint>child).constraint;
```
beaugogh: this also covers the variable constraint unwrapping?

beaugogh: just make sure we don't introduce old bugs back

PR #576, opened 2019-05-09
Glowing-bear docker image

ewelinagr:
- Add dockerfile and configs to build glowingbear image
- Add nginx config - currently only to serve the static config of gb, without ssl configuration (TODO)
- Add a first draft version of docker-compose script combining glowing-bear, transmart-api-server (+db), gb-backend (+db), transmart-packer - to be moved to a separate repository and split into multiple docker-compose scripts per component (UPDATE - moved to [glowing-bear-docker](https://github.com/thehyve/glowing-bear-docker/pull/13))


docker/config.template.json
```diff
@@ -0,0 +1,21 @@
+{
+  "api-url": "${TRANSMART_API_SERVER_URL}",
+  "api-version": "v2",
+  "app-version": "${GB_VERSION}",
+  "gb-backend-url": "${GB_BACKEND_URL}",
+  "doc-url": "https://glowingbear.app",
+  "enable-fractalis-analysis": false,
+  "autosave-subject-sets": false,
+  "show-observation-counts": true,
```
gijskant: maybe good to have this default to `false` as well?

docker/nginx.conf
```diff
@@ -0,0 +1,20 @@
+worker_processes 1;
+
+events { worker_connections 1024; }
+
+http {
+  include /etc/nginx/mime.types;
+
+  server {
+  	listen 80;
```
gijskant: Maybe good to expose this at a port `> 1024`, because otherwise this only works for root users and if you don't already have a webserver running.

docker/docker-compose.yml
```diff
@@ -0,0 +1,127 @@
+version: '3'
+services:
+  # ----- glowing-bear -----
+  glowing-bear:
+    container_name: glowing-bear
+    image: 'thehyve/glowing-bear'
+    ports:
+      - 80:80
```
gijskant: Maybe good to expose this at a port > 1024, because otherwise this only works for root users and if you don't already have a webserver running.

ewelinagr:
Thanks for the comments @gijskant, I added a short README, similar to the transmart-core/docker one.

PR #584, opened 2019-06-13
Define version only in package.json.

gijskant:


build.gradle
```diff
@@ -1,27 +1,18 @@
...
+
 }
 
-def versionFile = new File('src/app/config/config.default.json')
+def versionFile = new File('package.json')
 def versionObject = new JsonSlurper().parseText(versionFile.text)
 
 ext {
-  glowingBearVersion = versionObject['app-version']
-  nodeVersion = '11.10.0'
```
ewelinagr: are there some issues with version `11.10`?

ewelinagr:
there are still some `app-version` occurrences in the code that would be useful to remove, e.g.: [app.config.mock](https://github.com/thehyve/glowing-bear/blob/consistent-versioning/src/app/config/app.config.mock.ts#L15), [app config test](https://github.com/thehyve/glowing-bear/blob/consistent-versioning/src/app/config/app.config.spec.ts#L74) (probably to be changed to test the new way?), [docker config template](https://github.com/thehyve/glowing-bear/blob/consistent-versioning/docker/config.template.json#L3)

PR #585, opened 2019-06-14
 Fix parent nodes selection when importing variables for export 

ewelinagr:
There is an issue in PrimeNG library, when selecting tree nodes programmatically -  parent nodes are not marked as selected or partially selected. A consequence of this in GB is that it is really hard to find which variables were selected in the tree when importing variables in the export tab. 

Relate issue: [TMT-880](https://jira.thehyve.nl/browse/TMT-880)

### DIFF
--- a/src/app/modules/gb-side-panel-module/accordion-components/gb-variables/gb-variables.component.html
+++ b/src/app/modules/gb-side-panel-module/accordion-components/gb-variables/gb-variables.component.html
@@ -17,10 +17,10 @@
               (click)="importVariables()"
               type="button">
         <span>import</span>
-        <input [id]="fileElementId"
-               type="file"
-               style="display: none"/>
       </button>
+      <input [id]="variableFileElementId"
+             type="file"
+             style="display: none"/>
     </div>
   </div>
 
--- a/src/app/services/variable.service.ts
+++ b/src/app/services/variable.service.ts
@@ -17,6 +17,8 @@ import {CountItem} from '../models/aggregate-models/count-item';
 import {CountService} from './count.service';
 import {GbTreeNode} from '../models/tree-node-models/gb-tree-node';
 import {UIHelper} from '../utilities/ui-helper';
+import {TreeNode} from 'primeng/api';
+import {TreeNodeHelper} from '../utilities/tree-node-helper';
 
 @Injectable({
   providedIn: 'root'
@@ -251,23 +253,19 @@ export class VariableService {
       if (node) {
         const val = fields.length < 2 ? node[fields[0]] : (node[fields[0]] || {})[fields[1]];
         if (values.includes(val) && !this.selectedVariablesTree.includes(node)) {
-          this.selectedVariablesTree.push(node);
-          UIHelper.selectPrimeNgTreeParentNodesRecursively(node.parent, this.selectedVariablesTree);
+          TreeNodeHelper.addNodeToSelectedNodes(node, this.selectedVariablesTree);
+          TreeNodeHelper.updateParentNodesSelectionRecursively(node.parent, this.selectedVariablesTree);
         } else if (
           !isSelectionIncremental
           && !values.includes(val)
           && this.treeNodeService.isVariableNode(node)
           && this.selectedVariablesTree.includes(node)
         ) {
-          const index = this.selectedVariablesTree.indexOf(node);
-          this.selectedVariablesTree.splice(index, 1);
+          TreeNodeHelper.removeNodeFromSelectedNodes(node, this.selectedVariablesTree);
+          TreeNodeHelper.updateParentNodesSelectionRecursively(node.parent, this.selectedVariablesTree);
         }
         if (node['children']) {
-          node.children.forEach((child) => {
-            if (child) {
-              child.parent = node
-            }
-          });
+          TreeNodeHelper.updateParentForAllChildren(node);
           this.selectVariablesTreeByFields(node['children'], values, fields, isSelectionIncremental);
         }
       }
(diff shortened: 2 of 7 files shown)


src/app/services/variable.service.ts
```diff
@@ -249,10 +250,9 @@ export class VariableService {
     nodes.forEach((node: GbTreeNode) => {
       if (node) {
         const val = fields.length < 2 ? node[fields[0]] : (node[fields[0]] || {})[fields[1]];
-        if (
-          values.includes(val)
-          && !this.selectedVariablesTree.includes(node)) {
+        if (values.includes(val) && !this.selectedVariablesTree.includes(node)) {
           this.selectedVariablesTree.push(node);
```
gijskant: These two lines do: 1) select a node 2) select its ancestors recursively. Instead, the newly created function could select a node with its ancestors recursively.

src/app/utilities/ui-helper.ts
```diff
@@ -24,4 +26,23 @@ export class UIHelper {
       }
     }).bind(this), delay);
   }
+
+  /**
+   * The PrimeNg library does not handle selecting of a parent node
```
gijskant: Code documentation is most valuable if it at least describes what it does. Why it is implemented is also useful to note, but maybe as a second remark.
```javascript
/**
 * For a tree node, this function selects its ancestors (parent nodes) recursively. (General description, one-liner if possible)
 * 
 * Then a detailed description may follow, e.g., when this should be used, and when not. Why is this needed and is no standard library solution sufficient, preconditions on state or input, side effects, etc.
 * @param
 * @return
 * @throw
 */
```

src/app/utilities/ui-helper.ts
```diff
@@ -24,4 +26,23 @@ export class UIHelper {
...
+   * when selecting the node programmatically
+   * @param node parent of which has to be (partially) selected
+   * @param selectedNodes array of currently selecte nodes
+   */
+  public static selectPrimeNgParentTreeNode(node: TreeNode, selectedNodes: GbTreeNode[]) {
+    if (node.parent != null) {
```
gijskant: If it would be `selectPrimeNgTreeNodeAndAncestors(node: TreeNode, selectedNodes: GbTreeNode[])`, the structure would be more like:
```typescript
if (node === null) {
  return;
}
selectedNodes.push(node);
if (!node.children.every((child: TreeNode) => selectedNodes.includes(child))) {
  node.partialSelected = true;
}
selectPrimeNgTreeNodeAndAncestors(node.parent);
```

gijskant: Right?

gijskant: My main concern is that this function only uses `node.parent` and never `node` itself. You might as well pass `node.parent` to the function then.

gijskant:
```typescript
public static function recursivelySelectParentNodes(parent: TreeNode, selectedNodes: GBTreeNode[])`
  if (parent === null) {
    return;
  }
  if (parent.children.every((child: TreeNode) => selectedNodes.includes(child))) {
    selectedNodes.push(parent);
  } else {
    parent.partialSelected = true;
  }
  selectPrimeNgTreeNodeAndAncestors(parent.parent);
}
```

ewelinagr: thanks for the comments, I'll update this

ewelinagr:
I just noticed the import function is called twice for some reason, I will check why...

gijskant:
Is the import function still called twice or was that issue also resolved?

ewelinagr:
@gijskant It is still there and I discovered one more issue with the tree selection, that is already present on dev. It is an issue related to shared concepts. 

Steps to reproduce with our test data, e.g. on [glowingbear-dev](https://glowingbear-dev.thehyve.net)
1. Go to the export tab,
2. Deselect all variables,
3. Select _Vital Signs_
4. Deselect _Private Studies/SHARED_CONCEPTS_STUDY_C_PRIV/Vital Signs/Heart Rate_

Issue: Heart Rate variable is deselected, but _Vital Signs_ parent node stays selected.

Fixing in progress...

ewelinagr:
@gijskant Both issues described above should be fixed now.

src/app/modules/gb-side-panel-module/accordion-components/gb-variables/gb-variables.component.html
```diff
@@ -17,10 +17,10 @@
               (click)="importVariables()"
               type="button">
         <span>import</span>
-        <input [id]="fileElementId"
-               type="file"
-               style="display: none"/>
       </button>
+      <input [id]="variableFileElementId"
```
gijskant: This breaks functionality, because `variableFileElementId` is not defined.

gijskant: I cannot reproduce the twice importing of variables. Let's investigate tomorrow.

src/app/utilities/tree-node-helper.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * Function to partially select a tree node with the '-' icon.
+   *
+   * Partial selection means that only part of the children nodes is selected.
+   *
+   * @param node to be partially select
+   * @param selectedNodes current tree nodes selection
+   */
+  private static updateNodePartialSelection(node: GbTreeNode, selectedNodes: GbTreeNode[]) {
```
gijskant: Separating into functions is in general a good thing, but here, I think it's best to have the three cases together in one function: 1) all children selected 2) some children (partially) selected 3) no children (partially) selected. That is easier to read in my opinion.

PR #586, opened 2019-06-17
Make scroll-bar accessible for a large list of variables

ewelinagr:
Fixes [TMT-884](https://jira.thehyve.nl/browse/TMT-884)


gijskant:
Interestingly, now the scroll bar disappears at the bottom.

PR #587, opened 2019-06-18
 Disable the dimension selection dropdown after restoring a cohort 

ewelinagr:
Fixes [TMT-862](https://jira.thehyve.nl/browse/TMT-862) - when restoring a cohort, the dimension button should be locked.

Additionally, remove unused code in the cohort service.

### DIFF
--- a/src/app/services/cohort.service.ts
+++ b/src/app/services/cohort.service.ts
@@ -80,12 +80,6 @@ export class CohortService {
     this.updateCountsWithCurrentCohort();
   }
 
-  clearAll(): Promise<any> {
-    this.constraintService.clearCohortConstraint();
-    this.isDirty = true;
-    return this.updateCountsWithCurrentCohort();
-  }
-
   /**
    * ----------------------------- Update the current cohort panel -----------------------------
    */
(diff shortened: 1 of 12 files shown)


src/app/services/cohort.service.ts
```diff
@@ -80,12 +80,6 @@ export class CohortService {
     this.updateCountsWithCurrentCohort();
   }
 
-  clearAll(): Promise<any> {
```
gijskant: So, `clearAll` in `src/app/services/mocks/cohort.service.mock.ts` can also be removed?

gijskant:
Only works for the top level now. If I create a query "Select all biomaterials for patients with gender 'Female'", the dimension dropdown is disabled on the inner query. However, when saving and restoring that query, the dropdown is not disabled.
Screenshot example biomaterial query

src/app/services/cohort.service.ts
ewelinagr: it is removed from there as well

gijskant: ah, sorry, I missed that

ewelinagr:
thanks @gijskant, the problem you mentioned is fixed now.

gijskant:
The discrepancy now seems to be the other way around: when its is not disabled when dragging in (e.g., type from 100 cats), it is disabled after restore.

ewelinagr:
> The discrepancy now seems to be the other way around: when its is not disabled when dragging in (e.g., type from 100 cats), it is disabled after restore.

this is actually done on purpose, but let's discuss the ticket tomorrow then.

PR #590, opened 2019-06-20
Fix updating of cross table subject constraints

ewelinagr:
Cross table subject constraints were never updated - always set to True Constraints.
Fixes: [TMT-887](https://jira.thehyve.nl/browse/TMT-887)

### DIFF
--- a/src/app/services/cross-table.service.ts
+++ b/src/app/services/cross-table.service.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2017 - 2018  The Hyve B.V.
+ * Copyright 2017 - 2019  The Hyve B.V.
  *
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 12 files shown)


src/app/services/cross-table.service.ts
```diff
@@ -41,7 +42,8 @@ export class CrossTableService {
   private _crossTable: CrossTable;
   private _selectedConstraintCell: GbDraggableCellComponent;
 
-  constructor(private resourceService: ResourceService,
+  constructor(private constraintService: ConstraintService,
```
gijskant: Maybe also add a trigger to update the cohort selection when the cohort selection changes (just as Fractalis has)?
```typescript
this.cohortService.cohortsUpdated.asObservable()
      .subscribe((cohorts: Cohort[]) => {
        this.updateCells()
      });
```

src/app/services/cross-table.service.ts
```diff
@@ -181,6 +183,7 @@ export class CrossTableService {
       try {
         this.crossTable.rowHeaderConstraints = this.crossConstraints(this.rowConstraints);
         this.crossTable.columnHeaderConstraints = this.crossConstraints(this.columnConstraints);
+        this.crossTable.constraint = this.constraintService.cohortSelectionConstraint;
```
gijskant: Shouldn't this be the combination of all selected cohorts? And then also for exports? Looks like some of that code (in cohort service could use some refactoring...)

gijskant:
Works great! A bit like magic, so we may want to add spinners or some other progress indicators later, but that's a separate issue.
Could you fix the linting errors? (run `yarn lint` to run the report)

PR #597, opened 2019-07-09
Remove SSL support from the Docker image.

gijskant:


ewelinagr:
Should the image version be different though? 2.0.2 was already published with different nginx settings

PR #601, opened 2019-07-25
Change the description of negation on observation constraints.

gijskant:
Negation on observations constraints is complicated when using
nested subject dimension subselects. It does not trigger an
existential constraint (for this dimension element no observation
such and such exists), but it inverts the observation selection
(for this dimension an observation exists that does not match
these criteria).

### DIFF
--- a/src/app/modules/gb-cohort-selection-module/constraint-components/gb-constraint/gb-constraint.component.ts
+++ b/src/app/modules/gb-cohort-selection-module/constraint-components/gb-constraint/gb-constraint.component.ts
@@ -98,7 +98,7 @@ export class GbConstraintComponent implements OnInit {
   get observationBoxMessage(): string {
     let parentDimension = this.constraint.parentDimension;
     if (this.constraint.negated) {
-      return `for the ${parentDimension} there is NO observation:`;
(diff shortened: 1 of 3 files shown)


src/app/modules/gb-cohort-selection-module/constraint-components/gb-constraint/gb-constraint.component.ts
```diff
@@ -98,7 +98,7 @@ export class GbConstraintComponent implements OnInit {
   get observationBoxMessage(): string {
     let parentDimension = this.constraint.parentDimension;
     if (this.constraint.negated) {
-      return `for the ${parentDimension} there is NO observation:`;
+      return `for the ${parentDimension} there is an observation other than observations:`;
```
ewelinagr: one related cypress test: https://github.com/thehyve/glowing-bear/blob/dev/cypress/support/step_definitions/constraint-creation.js#L90

gijskant: Thanks!

gijskant: I updated it.

PR #606, opened 2019-09-03
Fix ensuring patient-level constraint for non-combination constraints

ewelinagr:
Fixes [TMT-987](https://jira.thehyve.nl/browse/TMT-987) - missing variables in variables panel, when selecting single cohort with concept constraint


src/tests/cohort.integration.spec.ts
```diff
@@ -223,6 +224,23 @@ describe('Integration test for cohort saving and restoring', () => {
       });
   });
 
+  it('should ensure patient level constraint before updating counts', () => {
```
gijskant: This kind of tests is a real anti-pattern. It tests whether a particular chain of function call is being done instead of testing whether certain input results in expected output.

ewelinagr: it's replaced with unit tests now.

PR #612, opened 2019-09-17
Cross table improvements

gijskant:
- Show scrollbar when cross table size exceeds panel width.
- Only enable adding cross table when none is present.
- Remove 'clear cache' button when Fractalis is not enabled.
- Do not change row constraints when loading
- Update cross table layout.

Fixes [TMT-1007](https://jira.thehyve.nl/browse/TMT-1007).


src/app/modules/gb-analysis-module/cross-table-components/gb-cross-table/gb-cross-table.component.ts
```diff
@@ -23,6 +24,8 @@ import {FormatHelper} from '../../../../utilities/format-helper';
 })
 export class GbCrossTableComponent implements OnInit {
 
+  AxisType = AxisType;
```
ewelinagr: Why is this needed?

gijskant: Otherwise you cannot use the type in the html template.

PR #613, opened 2019-09-19
Check server status at startup.

gijskant:


src/app/app.component.html
```diff
@@ -1,9 +1,44 @@
...
+        <p>The configuration is not loaded correctly</p>
+        <code *ngIf="configError">
+          {{configError}}
+        </code>
+      </div>
+    </div>
+
+    <div class="gb-page" *ngIf="serverDown">
+      <div class="gb-message-box gb-error">
+        The server is down.
```
ewelinagr: Should it be only for transmart? Or gb-backend as well? It also has the availability endpoint.

gijskant: Indeed, that would be a great addition. I'll create an issue for that.

PR #619, opened 2019-10-03
Fix data table mapper and rendering.

gijskant:


src/app/modules/gb-export-module/data-table-components/gb-data-table-grid/gb-data-table-grid.component.html
```diff
@@ -5,22 +5,11 @@
...
-        </tr>
-      </ng-template>
-
       <!-- table body -->
       <ng-template pTemplate="body" let-row let-columns="columns">
-        <tr>
-          <td *ngFor="let col of columns" class="ui-resizable-column">
+        <!-- Header row -->
+        <tr *ngIf="row.isHeaderRow">
+          <th *ngFor="let col of columns" class="table-header">
```
ewelinagr: Maybe better to keep `pResizableColumn` option? Also, with many columns it's unreadable now (example below), maybe some min-width of columns would be useful? 
image

gijskant: I now changed the data table to auto layout. That improves the rendering of tables with many columns. The problem with the resizable columns is that is has quite stange behaviour when resizing.

PR #624, opened 2019-11-26
Changes related to a new upcoming fractalis version

ewelinagr:
To be merged after new releases of https://github.com/thehyve/Fractal.js and https://github.com/thehyve/Fractalis


ewelinagr:
Failing tests related to the fractal.js version. They are passing with version from https://github.com/thehyve/Fractal.js/pull/4

ewelinagr:
I tried async/await approach for fetching the token synchronously in [getAuth](https://github.com/thehyve/glowing-bear/blob/fractalis/src/app/services/fractalis.service.ts#L100), but this would require adding async/await also in Fractal.js.

PR #627, opened 2019-12-12
Replace fractalis handler and datasource url with datasource service

ewelinagr:


gijskant:
The test do pass with the updated fractalis package, I suppose?

ewelinagr:
yes, I tested locally with our version of Fractal.js and they pass

PR #631, opened 2020-02-18
Upgrade cypress, simplify e2e tests, remove transmart auth option.

gijskant:


ewelinagr:
The `authentication-service-type` could be probably removed from all the places, e.g.: [app.config.ts#L102](https://github.com/thehyve/glowing-bear/blob/e2e-tests/src/app/config/app.config.ts#L102), [README](https://github.com/thehyve/glowing-bear/blob/e2e-tests/README.md#configuration), [config.default.json#L10](https://github.com/thehyve/glowing-bear/blob/e2e-tests/src/app/config/config.default.json#L10) etc...
Maybe also the puppet module?

gijskant:
I agree, I will have another round of cleanup...

PR #648, opened 2021-04-22
Release version 2.0.14.

gijskant:
Changes the way the `KEYCLOAK_SERVER_URL` variable is used in the Docker image to set `oidc-server-url`. Keycloak does not always use a `/auth` path prefix. When it does, it needs to be included in the variable, e.g., `KEYCLOAK_SERVER_URL=https://keycloak.example.com/auth`. Previously, the `/auth` prefix was appended automatically.


gijskant:
@ewelinagr Do you think it is okay to release this one? Then I can update glowing-bear-docker and all components are up-to-date.

ewelinajozwik (approved):

